PR #11, opened 2016-01-28
Add header to extend server for arduinojson

sticilface:


src/AsyncJson.h
```diff
@@ -0,0 +1,76 @@
...
+    size_t _to_skip;
+    size_t _to_write;
+    size_t _pos;
+  public:
+    ChunkPrint(uint8_t* destination, size_t from, size_t len)
+      : _destination(destination), _to_skip(from), _to_write(len), _pos{0} {}
+    size_t write(uint8_t c){
+      if (_to_skip > 0) {
+        _to_skip--;
+        return 0;
```
me-no-dev: I wander if it should return 1 here? If ArduinoJson would have stopped sending when it returns 0 then time would have been less. Maybe we should look at that and fix it there :)

PR #18, opened 2016-04-06
Gracefully handle out of memory

andig:
This strategy seems to work for me and seems clearer and simpler than some sort of backoff algorithm.

I'm assuming that returning 0 will keep the request open. Could you check if https://github.com/me-no-dev/ESPAsyncWebServer/compare/master...andig:fix-lowmem?expand=1#diff-2bfef7483934cf6c10c5c03a8553d5b0R273 is correct? I'm only guessing here according to the rest of the code.

Without this fix I'm into crashes right away when serving multiple static files under load.

### DIFF
--- a/src/WebResponses.cpp
+++ b/src/WebResponses.cpp
@@ -267,6 +267,10 @@ size_t AsyncAbstractResponse::_ack(AsyncWebServerRequest *request, size_t len, u
       outLen = (remaining > space)?space:remaining;
     }
     uint8_t *buf = (uint8_t *)malloc(outLen);
+    if (!buf) {
+      // os_printf("_ack malloc %d failed\n", outLen);
+      return 0;
+    }
 
     if(_chunked){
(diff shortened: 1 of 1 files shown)


me-no-dev:
should not be in wait ack as the previous packet is acked. the state should not need change and the poll should send the remaining data (given that it can malloc)

me-no-dev:
doing what you propose will kill the connection and I do not think that that is necessary :)
the onPoll will be called in some milliseconds and memory might be available then, so I think that the lib should let the connection wait for more data :)

andig:
So what should I change- just remove the ack or try to allocate freemem/2? Feel free to ignore the pr and take another approach, but this malloc needs a fix ;)

andig:
Would it be sufficient to just return 0?

me-no-dev:
yes. returning 0 here should be enough.

andig:
Fixed.

PR #26, opened 2016-05-06
Prevent Buffer Overflow

hallard:
Not sure buffer received can handle one more byte for '\0' in case of WS_TEXT data for C string conversion
Not the best optimization code but works.

### DIFF
--- a/src/AsyncWebSocket.cpp
+++ b/src/AsyncWebSocket.cpp
@@ -495,6 +495,15 @@ void AsyncWebSocketClient::text(const char * message, size_t len){
 void AsyncWebSocketClient::text(const char * message){
   text(message, strlen(message));
 }
+void AsyncWebSocketClient::text(uint8_t * message, size_t len){
+  text((const char *)message, len);
+}
+void AsyncWebSocketClient::text(char * message){
+  text(message, strlen(message));
+}
+void AsyncWebSocketClient::text(String &message){
+  text(message.c_str(), message.length());
+}
 void AsyncWebSocketClient::text(const __FlashStringHelper *data){
   PGM_P p = reinterpret_cast<PGM_P>(data);
   size_t n = 0;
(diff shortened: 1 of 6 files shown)


src/AsyncWebSocket.cpp
```diff
@@ -725,6 +725,22 @@ void AsyncWebSocket::text(uint32_t id, uint8_t * message, size_t len){
 void AsyncWebSocket::text(uint32_t id, char * message){
   text(id, message, strlen(message));
 }
+void AsyncWebSocket::text(uint32_t id, const __FlashStringHelper *data) {
```
me-no-dev: those method should be added to the AsyncWebSocketClient and pointed to like the other text and binary methods.

src/AsyncWebSocket.cpp
```diff
@@ -725,6 +749,21 @@ void AsyncWebSocket::text(uint32_t id, uint8_t * message, size_t len){
...
+      n += 1;
+  }
+  char * message = (char*) malloc(n+1);
+  if(message){
+    for(size_t b=0; b<n; b++)
+      message[b] = pgm_read_byte(p++);
+    message[n] = 0;
+    text(id, message, n);
+  }
+}
```
me-no-dev: maybe this instead:
```cpp
 void AsyncWebSocket::text(uint32_t id, const __FlashStringHelper *message){
   AsyncWebSocketClient * c = client(id);
   if(c != NULL)
     c->text(message);
 }
 
 void AsyncWebSocket::textAll(const __FlashStringHelper *message){
   AsyncWebSocketClient * c = _clients;
   while(c != NULL){
     if(c->status() == WS_CONNECTED)
       c->text(message);
     c = c->next;
   }
 }
```
same should be done for binary and binaryAll

me-no-dev:
@hallard you around?

hallard:
Yeah in about 1h

me-no-dev:
I would like to merge this when you have time :)

hallard:
no problem, I'm now there and on gitter

me-no-dev:
I would love it if you make the changes I outlined above :) and I will merge it.
The reason I ask for them is because the methods in AsyncWebSocket should only proxy to the client's methods and not actually implement them :) you can see the rest of the methods done that way, with the exception of printf

PR #36, opened 2016-06-09
  This handler will serve a 302 response to a client request for a SP…

sticilface:
Serves a 302 with location if the request comes from the STA side of the ESP network. If the request comes from the AP side then it serves the file from SPIFFS.

PR #37, opened 2016-06-12
Fix bug in AsyncStaticWebHandler

hagai-shatz:
Add support for http redirection (302) and http not modified (304) to
reduce the load the server.
server.redirect(“url”, “location”, exclude-ip) will respond with 302 to
redirect the browser to a different url, this is useful for backward
compatibility and to redirect call to CDN when not no AP mode.
server.serveStatic has a new optional parameter to get the
Last-Modified date for all files serve for this location, when the
browser request have the same If-Modified-Since header value, the
server respond with 304 code instead of serving the file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,6 +397,26 @@ response->setLength();
 request->send(response);
 ```
 
+### FileFallBackHandler
+Example provided by [@sticilface](https://github.com/sticilface)
+
+This handler is useful for serving content from a CDN when the ESP is connected to a wifi 
+network, but falling back to local copies of the file stored in SPIFFS when the ESP is in 
+AP mode and the client does not have internet access.  It will work when both AP mode and 
+STA mode are active.  It works by returning 302 HTTP code, with a Location header that 
+you specify.  It is much quicker than requiring the ESP to handle all the files. 
+```cpp
+#include "FileFallbackHandler.h" //  include this in the sketch. 
+
+server.addHandler( new FileFallbackHandler(SPIFFS, "/path_to_SPIFFS_file", "/uri", "url_to_forward",  "optional_cache_control_header"));
+
+//  These three lines will serve all the jquery requirements from SPIFFS (if they are there) in AP mode, but forward the URL to CDN if not. 
+server.addHandler( new FileFallbackHandler(_fs, "/jquery/jqm1.4.5.css", "/jquery/jqm1.4.5.css",   "http://code.jquery.com/mobile/1.4.5/jquery.mobile-1.4.5.min.css",  "max-age=86400"));
+server.addHandler( new FileFallbackHandler(_fs, "/jquery/jq1.11.1.js" , "/jquery/jq1.11.1.js" ,   "http://code.jquery.com/jquery-1.11.1.min.js",  "max-age=86400"));
+server.addHandler( new FileFallbackHandler(_fs, "/jquery/jqm1.4.5.js" , "/jquery/jqm1.4.5.js" ,   "http://code.jquery.com/mobile/1.4.5/jquery.mobile-1.4.5.min.js",  "max-age=86400"));
+```
+
+
 ## Bad Responses
 Some responses are implemented, but you should not use them, because they do not conform to HTTP.
 The following example will lead to unclean close of the connection and more time wasted
--- a/README.md
+++ b/README.md
@@ -625,3 +625,36 @@ void setup(){
 
 void loop(){}
 ```
+
+### Methods for controlling websocket connections
+
+```arduino
+  // Disable client connections if it was activated
+  if ( ws.enabled() )
+    ws.enable(false);
+
+  // enable client connections if it was disabled
+  if ( !ws.enabled() )
+    ws.enable(true);
+```
+
+Example of OTA code 
+
+```arduino
+  // OTA callbacks
+  ArduinoOTA.onStart([]() {
+    // Clean SPIFFS
+    SPIFFS.end();
+
+    // Disable client connections    
+    ws.enable(false);
+
+    // Advertise connected clients what's going on
+    ws.textAll("OTA Update Started");
+
+    // Close them
+    ws.closeAll();
+
+  });
+
+```
--- a/src/AsyncWebSocket.cpp
+++ b/src/AsyncWebSocket.cpp
@@ -499,7 +499,7 @@ size_t AsyncWebSocketClient::printf_P(PGM_P formatP, ...) {
 #else
   size_t len = vsnprintf(NULL, 0, format, arg)+1;
 #endif
-  size_t fmtLen = strlen_P(formatP); 
+  size_t fmtLen = strlen_P(formatP);
   format = (char*)calloc(fmtLen+1, sizeof(char));
   if ( format ) {
     strcpy_P(format, formatP);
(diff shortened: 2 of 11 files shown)


hagai-shatz:
I believe this is better implementation to redirect (302) responses than [FileFallbackHandler](https://github.com/me-no-dev/ESPAsyncWebServer/pull/36) since this solution does not duplicate the code in AsyncStaticWebHandler, instead it built on the order the handlers are added to the server and if the redirect handler does not redirect, the static handler will serve the file (if exists). See example in readme.

README.md
```diff
@@ -416,6 +416,38 @@ server.addHandler( new FileFallbackHandler(_fs, "/jquery/jq1.11.1.js" , "/jquery
...
+
+```
+
+### HTTP Redirect (302)
+Return HTTP 302 response to redirect the client to a different location.
+An optional IPAddress can specify redirection exclusions, for example, exclude redirection to requests to the AP address.
+```cpp
+// Redirect all requests
+server.redirect("/root/file.ext", "/other-location/other-file.ext");
+// Redirect requests not for the AP address
```
me-no-dev: the redirect by IP for files seems to propose the same functionality as the FileFallbackHandler documented above.
You can see that it also has configuration for which file system to use, because the server can run from SD or other storage.
Another note on redirecting to local files like this is that you require another request to be made and parsed by the server, while you can serve the local file with the first one.
I agree that redirecting to URLs is a good idea, but let's not document that as redirecting to files like this (for CDN) and use the FileFallbackHandler for handling that and document other use cases where redirection makes more sense and does not cause extra requests to be sent to the ESP if they can be avoided :)
FileFallbackHandler could be maybe better exposed through server.serveCDN().
I feel that I need to brainstorm a bit more on the redirect thing. I'm all up for the serveStatic changes for sure, though maybe you should think of a way to update the modification date after the sketch is started (and the file is updated)

hagai-shatz: Indeed, this is same functionality as in FileFallbackHandler but in a more generic design: the redirect is general redirect regardless of file serving. If there is no redirect (for example, call to the AP IP address), the server will fallback to whatever handler can handle the request (this can be static file from SPIFFS or SD card, or even a callback handler. I agree that it is better to avoid redirection altogether and design the solution to perform the request to the appropriate url to begin with.
As for redirecting to CDN - this solution assume that the browser have internet access when request is done on STA mode, however, this might not be true. Another approach might be to serve different static file based on the client local IP, that is, for requests done on AP we serve a file that link to the local resources, for requests done on STA we serve a different file that link to the resources on the CDN. This will remove the need for the ESP to redirect.
Last thing - I agree that it will be good to set the modified date without a change to the sketch, one option is to load the modified date from a config file and update this file when you upload new files to serve, however, this will require a restart to the ESP in order to do the setup again to setup the server with the new date. We need some API to set the modified date for the handler after it was added to avoid restart.

me-no-dev: > for requests done on AP we serve a file that link to the local resources, for requests done on STA we serve a different file that link to the resources on the CDN

this is exactly what the fallback handler does. If AP client, serve the file, else 302 to the CDN.
As for wether the ESP has internet on the STA side, that is something I imagine will be known for the particular implementation and not offer CDN at all if it is expected to be offline.

BTW do you mind coming to gitter so we can speed up the communication and exchange more ideas :)

hagai-shatz: What the FileFallbackHandler does is a combination of redirect or serve file, not the best since there is a request to the ESP. The redirect functionality in my pull request does the same but in a more generic way - redirect to CDN if not on AP side, otherwise look for another handler to serve the request.

Using FileFallbackHandler:
server.addHandler( new FileFallbackHandler(SPIFFS, "/jquery/jq1.11.1.js" , "/jquery/jq1.11.1.js" ,   "http://code.jquery.com/jquery-1.11.1.min.js",  "max-age=86400"));

Using my code:
server.redirect("/jquery/jq1.11.1.js", "http://code.jquery.com/jquery-1.11.1.min.js", WiFi.softAPIP());
server.serveStatic("/jquery/jq1.11.1.js", SPIFFS, "/jquery/jq1.11.1.js", "max-age=86400");

However, the serve static can serve folder instead of individual file:
server.redirect("/jquery/jqm1.4.5.css", "http://code.jquery.com/mobile/1.4.5/jquery.mobile-1.4.5.min.css", WiFi.softAPIP());
server.redirect("/jquery/jq1.11.1.js", "http://code.jquery.com/jquery-1.11.1.min.js", WiFi.softAPIP());
server.redirect("/jquery/jqm1.4.5.js", "http://code.jquery.com/mobile/1.4.5/jquery.mobile-1.4.5.min.js", WiFi.softAPIP());
server.serveStatic("/jquery", SPIFFS, "/jquery", "max-age=86400");

Both solutions are based on html page with the following in the header:
    <link rel="stylesheet" href="jquery/jqm1.4.5.css" />
    <script src="jquery/jq1.11.1.js"></script>
    <script src="jquery/jqm1.4.5.js"></script>

What I'm suggesting is to serve 2 different html index pages, one when index.htm is request on AP with the above header, and another when index.htm is requested on STA with header that link directly to the CDN to skip the request to the ESP and the redirection.
Something like:
server.serveStaticOnAP("/", SPIFFS, "/ap");
serverServeStatic("/", SPIFFS, "/www");
"/ap" directory only include the file "index.htm" with header links to "/jquery/..". "/www" directory include all assets including "index.htm" with header links to the CDN.

me-no-dev: oh oh, I think I'm starting to grasp the idea :) I think I missed the part where you let another handler take care of the request.
As for a different index or any other page, I think the FileFallbackHandler needs to be updated to handle two files, like:
```cpp
server.addHandler(new FileFallbackHandler(SPIFFS, "/" , "/index-ap.html" , "/index-sta.html", "max-age=86400"));
```
the other way is to actually implement some sort of rewrite functionality, that can rewrite the request's url and host based on the request (before the handlers are checked if they can handle it)

me-no-dev:
@sticilface would you please join this discussion, so we can better implement the new functionality and maybe get rid of FileFallbackHandler or have it interact differently.

Here is what I propose:
```cpp
//serve file from CDN if STA client
server.serveCDN("/jquery.js", "http://example.com/jquery9.9.9.js");
//else serve using serveStatic
server.serveStatic("/", SPIFFS, "/");
server.serveStatic("/", SPIFFS, "/", "max-age=86400");
//rethink how to deal better with this and be able to change the timestamp on update 
server.serveStatic("/", SPIFFS, "/", NULL, "Wed, 08 Jun 2016 22:00:00 GMT");
//redirect actually to be added to the request and be used like any other response
request->redirect("some-url");
//create rewrite rules that are executed before a handler is assigned to the response
server.rewrite("/index.html", "/index.htm");//always
server.rewrite("/index.htm", "/ap.htm", [](AsyncWebServerRequest *request) -> bool {
  if(request->localIP() == WiFi.softAPIP())
    return true;//rewrite
  return false;//leave as is
});
```
or something in this context

me-no-dev:
another idea for the 304 response
how about we have that separate in the form of:
```cpp
server.cacheFile("/index.htm", "Wed, 08 Jun 2016 22:00:00 GMT");
```
calling this again later on will update the given file's modified date. Calling it without the date should set the date to the sketch build date.
That way file source and file modified dates are not tied together, and can be easily updated.

hallard:
Guys,
Love this one !!!!!!

hagai-shatz:
I'm very sorry but I've messed up my master branch.
This is no longer a pull request for "HTTP 302 and 304 Support", rename to "Fix bug in AsyncStaticWebHandler"

Issue #27, opened 2016-05-08
serveStatic("/" ...) results in HTTP 500


mfalkvidd:
readme.md says:
```
  // attach filesystem root at URL /fs 
  server.serveStatic("/fs", SPIFFS, "/");
```
I wanted to serve the root uri from /html/ on SPIFFS, so I used
`  server.serveStatic("/", SPIFFS, "/html/");`
but this results in HTTP 500. The following syntax works though:
`  server.serveStatic("", SPIFFS, "/html/");`

vitotai:
I might have the same issue as you.
I solved it by modifying WebHandlers.cpp
String AsyncStaticWebHandler::_getPath(AsyncWebServerRequest *request)

    if (!baserequestUrl.length()) {
      baserequestUrl += "/";
    }

change to 
	
    if (!baserequestUrl.length()) {
    	if( _uri != "/")
      		baserequestUrl += "/";
    }

ccoenen:
If anyone else lands here with SPIFFS / serveStatic and HTTP 500 errors, one thing that was not immediately obvious to me: you have to call `SPIFFS.begin()` before you can server anything successfully.

PR #41, opened 2016-06-15
Fix path problems in static handler and improve performance.

hagai-shatz:
This fix the problems with specifying root path “/“ to serve or to map (issue #27).
It also improve performance by reducing the number of calls to _fs.exists in most cases. The handler also maintain statistic on the last 8 files that found if they are gzip and try to find a “.gz” files first if more files are server gripped.

### DIFF
--- a/src/WebHandlers.cpp
+++ b/src/WebHandlers.cpp
@@ -23,66 +23,81 @@
 
 bool AsyncStaticWebHandler::canHandle(AsyncWebServerRequest *request)
 {
-  if (request->method() != HTTP_GET) {
-    return false;
-  }
-  if ((_isFile && request->url() != _uri) ) {
-    return false;
-  }
-  //  if the root of the request matches the _uri then it checks to see if there is a file it can handle. 
-  if (request->url().startsWith(_uri)) {
-    String path = _getPath(request);
-    if (_fs.exists(path) || _fs.exists(path + ".gz")) {
-       DEBUGF("[AsyncStaticWebHandler::canHandle] TRUE\n");
-      return true;
-    }
+  if (request->method() == HTTP_GET &&
+      request->url().startsWith(_uri) &&
+      _getPath(request, true).length()) {
+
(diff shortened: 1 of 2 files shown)


src/WebHandlers.cpp
```diff
@@ -23,73 +23,86 @@
...
 
-  DEBUGF(" final path = %s\n", path.c_str());
-  DEBUGF("[AsyncStaticWebHandler::_getPath] END\n\n");
+  bool found = fileFound || gzipFound;
+
+  if (withStats && found) {
+    _gzipStats = (_gzipStats << 1) + gzipFound ? 1 : 0;
+    _fileStats = (_fileStats << 1) + fileFound ? 1 : 0;
+    _gzipFirst = _countBits(_gzipStats) > _countBits(_fileStats);
+  }
```
me-no-dev: could you please explain what that does? I'm a bit lost at how is ```_gzipFirst``` calculated :)

hagai-shatz: _gzipStats and _fileStats are shifted left every time we find a file, the LSB is set for _gzipStats if we have found a gzip file, or to _fileStats if we found a regular file, but never to both of them.
For example: after serving gzip, file, gzip, gzip, file, gzip. _gzipStats = 0b00101101 and _fileStats = 0b00010010. We then count the number of set bits in each and if we served more gzip files _gzipFirst is set so the code will do exists(path+'.gz') before exists(path) to avoid the extra call in most cases.
A faster option is to use just one byte _gzipStats and set _gzipFirst =  _countBits(_gzipStats) > 4, but this will require 4 to 8 files to warm up:
`
+  if (withStats && found) {
+    _gzipStats = (_gzipStats << 1) + gzipFound ? 1 : 0;
+    _gzipFirst = _countBits(_gzipStats) > 4;
+  }
`

me-no-dev: ok, interesting :) let's give it a go

PR #42, opened 2016-06-17
change quotes from ' ' to " " for filename header

everslick:
Some Browsers don't like single ' around parameters in headers and store them as part of the value.

### DIFF
--- a/src/WebResponses.cpp
+++ b/src/WebResponses.cpp
@@ -365,10 +365,10 @@ AsyncFileResponse::AsyncFileResponse(FS &fs, String path, String contentType, bo
 
   if(download) {
     // set filename and force download
-    snprintf(buf, sizeof (buf), "attachment; filename='%s'", filename);
+    snprintf(buf, sizeof (buf), "attachment; filename=\"%s\", filename);
   } else {
     // set filename and force rendering
-    snprintf(buf, sizeof (buf), "inline; filename='%s'", filename);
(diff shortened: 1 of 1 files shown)


src/WebResponses.cpp
```diff
@@ -365,10 +365,10 @@ AsyncFileResponse::AsyncFileResponse(FS &fs, String path, String contentType, bo
 
   if(download) {
     // set filename and force download
-    snprintf(buf, sizeof (buf), "attachment; filename='%s'", filename);
+    snprintf(buf, sizeof (buf), "attachment; filename=\"%s\", filename);
```
me-no-dev: missing quote :)

src/WebResponses.cpp
```diff
@@ -365,10 +365,10 @@ AsyncFileResponse::AsyncFileResponse(FS &fs, String path, String contentType, bo
...
   } else {
     // set filename and force rendering
-    snprintf(buf, sizeof (buf), "inline; filename='%s'", filename);
+    snprintf(buf, sizeof (buf), "inline; filename=\"%s\"", filename);
   }
   addHeader("Content-Disposition", buf);
```
me-no-dev: should be changed in the method below as well

src/WebResponses.cpp
everslick: argl, i really should not use githubs built in editor for PRs. :-D

PR #45, opened 2016-06-17
Bug fix

hagai-shatz:
Fix a bug to serve a file in directory (e.g. serverStatic("/", SPIFFS, "/index.html");


andig:
whoa, 11 commits?

hagai-shatz:
It is a dirty branch I have. I must be using GIT in the wrong way. There is only a small change with the latest master.

andig:
you can git reset to before your changes, stash them and the update your branch before popping the stash again

hagai-shatz:
Thanks for the tip. Next time.

andig:
I'm wondering- given the error potential and complexity of the code- if this is a good fix for the original issue.

hagai-shatz:
I think it is. I've tested it myself. _isDir is a hint to improve performance - avoid looking for the directory as it was a file just to fail and then add the default 'index.htm' to it and search again. The problem was that I was too optimistic setting this flag to true.

andig:
Confirmed wfm, thank you!

PR #47, opened 2016-06-21
New - rewrite, filters and last-modified

hagai-shatz:
New - server.rewrite() to rewrite the request url with optional get parameters injection.
New - rewrite.setFilter() and handler.setFilter() to specify a filter callback for more control on when to include them.
New - static file handler can be set with Last-Modified header value to support Not-Modified (304) response instead of serving the actual file.


src/ESPAsyncWebServer.h
```diff
@@ -286,31 +342,41 @@ typedef std::function<void(AsyncWebServerRequest *request, uint8_t *data, size_t
...
 
     void begin();
-    void addHandler(AsyncWebHandler* handler);
 
-    void on(const char* uri, ArRequestHandlerFunction onRequest);
-    void on(const char* uri, WebRequestMethod method, ArRequestHandlerFunction onRequest);
-    void on(const char* uri, WebRequestMethod method, ArRequestHandlerFunction onRequest, ArUploadHandlerFunction onUpload);
-    void on(const char* uri, WebRequestMethod method, ArRequestHandlerFunction onRequest, ArUploadHandlerFunction onUpload, ArBodyHandlerFunction onBody);
+    AsyncWebRewrite& addRewrite(AsyncWebRewrite* rewrite);
+    void clearRewrites(void);
```
me-no-dev: lets instead have 
```cpp
bool removeHandler(AsyncWebHandler* handler);
bool removeHandler(const char * url);
```
Same for the handlers

hagai-shatz: This may be okay for `bool removeHandler(AsyncWebHandler* handler);` but since we have filters, you might have several rewrite with the same url.

PR #70, opened 2016-08-19
added lower case string for 'Host'

jeppo77:
The 'request' npm package generates headers with "host" i.s.o. "Host". 
Not sure who is wrong, but had to change WebRequest.cpp for this.


me-no-dev:
I know about this issue and it's not related just to the host header :) i'll rework it hopefully next week

jeppo77:
ok great thanks!

me-no-dev:
should be resolved in the above commit

PR #84, opened 2016-10-20
fix & in param mistake

layet:
urlDecode must be called only on param value after parsing params to array

### DIFF
--- a/src/WebRequest.cpp
+++ b/src/WebRequest.cpp
@@ -219,7 +219,7 @@ void AsyncWebServerRequest::_addGetParams(String params){
     if (equal < 0 || equal > end) equal = end;
     String name = params.substring(start, equal);
     String value = equal + 1 < end ? params.substring(equal + 1, end) : String();
-    _addParam(new AsyncWebParameter(name, value));
+    _addParam(new AsyncWebParameter(name, urlDecode(value)));
     start = end + 1;
   }
 }
@@ -248,7 +248,7 @@ bool AsyncWebServerRequest::_parseReqHead(){
     _method = HTTP_OPTIONS;
   }
 
-  u = urlDecode(u);
+  //u = urlDecode(u);
   String g = String();
   index = u.indexOf('?');
   if(index > 0){


me-no-dev:
I do not agree ;) used to be that way but got many reports on the opposite

Adam5Wu:
It seems @layet is partially correct, that url decoding has to happen after the param spliting.
otherwise there is no way to send '&' in a query to the server
However, url decode should also happen on the non-query part as well, otherwise one cannot visit a path with space, since the handler would be looking for a directory/file with %20 in the name...

me-no-dev:
> there is no way to send '&' in a query to the server

did not quite get that

Adam5Wu:
Say you have a form, which allow user to enter text and submit as get request.
User enters R&B in the music field
It got encoded as
GET /url?music=R%26B
If url decoding was done before param processing, then it becomes
GET /url?music=R&B
Then got split into two params:
music=R
B=

Adam5Wu:
Similarly, but more rare, some one may create a form with some key contains & or = in it.
If urldecode done too early, param parsing also results in error

So it seems ir should be done only at last level, when no further processing is needed.
That is, on each of url path, key and value, separately.

me-no-dev:
I'm trying to remember what exactly was the reason behind that. I remember clearly having it the other way around (as requested here) and changing to the way it is now, because something was broken. Just not remembering what was broken and the case that breaks it.
Before committing to one way or another again, I would like all options (POST and GET variants) exhaustively tested.

Adam5Wu:
The current PR is incomplete, because the url path and keys are not url decoded

Adam5Wu:
For example, GET /music/R%26B/playlist
Will not match handler /music/R&B
Unless the path is urldecoded

me-no-dev:
isn't there a specification to go by, If we adhere to specs, it's good enough. There must be some algorithm implemented.

Adam5Wu:
Unfortunately, because this procedure is highly coupled with the parsing workflow on server side, there is probably no packaged one-call solution...

me-no-dev:
if we make a few different forms and watch the output it can be done ;) with data, patterns emerge :D

Adam5Wu:
https://tools.ietf.org/html/rfc3986 Page 22
According to RFC, we cannot even urldecode path as a whole
Each segment must be splited first, then urldecode

Adam5Wu:
I guess there is no need to go nuts and implement full specification compliance. 
As you said probably a couple trial and error and works for most common case is good enough.😊

PR #91, opened 2016-11-11
Resolved conflicted HTTP_[...] redeclaration 

nut-code-monkey:
Fix for issue https://github.com/marvinroger/homie-esp8266/issues/120 when ESPAsyncWebServer.h and ESP8266WebServer.h included

### DIFF
--- a/src/AsyncEventSource.cpp
+++ b/src/AsyncEventSource.cpp
@@ -253,7 +253,7 @@ size_t AsyncEventSource::count(){
 }
 
 bool AsyncEventSource::canHandle(AsyncWebServerRequest *request){
-  if(request->method() != HTTP_GET || !request->url().equals(_url))
+  if(request->method() != WebRequest::Method::HTTP_GET || !request->url().equals(_url))
     return false;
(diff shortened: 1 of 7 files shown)


me-no-dev:
I was not aware of this issue :) never tried to run both as well. Maybe we can find a better solution and use the same enum as the regular WebServer

etiquet:
hello, any chance the pull request succeed ?

src/AsyncEventSource.cpp
```diff
@@ -253,7 +253,7 @@ size_t AsyncEventSource::count(){
 }
 
 bool AsyncEventSource::canHandle(AsyncWebServerRequest *request){
-  if(request->method() != HTTP_GET || !request->url().equals(_url))
+  if(request->method() != WebRequest::Method::HTTP_GET || !request->url().equals(_url))
```
me-no-dev: sorry have not looked in a long time, but how about it is renamed to WebMethod instead of WebRequest::Method

me-no-dev:
This PR needs to be synced with master in order to merge. Please do read my other comment too.

PR #93, opened 2016-11-16
Reset all handlers. This allowing to use same server few times

nut-code-monkey:
E.g. for WiFi manager when hardware sepups WiFi connection and as HTTP server during hardware life cycle

### DIFF
--- a/README.md
+++ b/README.md
@@ -807,7 +807,7 @@ server.on("/scan", HTTP_GET, [](AsyncWebServerRequest *request){
 });
 ```
 
-### Remove handlers and writers
+### Remove handlers and rewriters
 
 Server goes through handlers in same order as they were added. You can't simple add handler with same path to override them.
 To remove handler:
@@ -819,9 +819,8 @@ auto handler = server.on("/some/path", [](AsyncWebServerRequest *request){
 // when you don't need handler anymore remove it
 server.removeHandler(&handler);
 
-// same with writers
-server.removeWriter(&someWriter);
-
+// same with rewriters
+server.removeRewrite(&someRewriter);
 
 server.onNotFound([](AsyncWebServerRequest *request){
 request->send(404);
--- a/README.md
+++ b/README.md
@@ -807,6 +807,31 @@ server.on("/scan", HTTP_GET, [](AsyncWebServerRequest *request){
 });
 ```
 
+### Reset particular handler
+
+Server goes through handlers in same order as they were added. You can't simple add handler with same path to override them.
+To remove handler:
+```arduino
+// save callback for particular URL path
+auto callback = server.on("/some/path", [](AsyncWebServerRequest *request){
+  //do something useful
+});
+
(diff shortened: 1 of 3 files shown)


me-no-dev:
I would love it if you add a few words in README about that feature :)

nut-code-monkey:
Hope last changes is more appropriate. Now method void reset() will accept AsyncCallbackWebHandler returned from .on(...). In that case user can to know anything about internal structure of Web Handler. And method could reset internal _catchAllHandler and user 'Handlers'.

me-no-dev:
This is more of what I imagined :)
```cpp
void AsyncWebServer::reset(){
  while(_rewrites != NULL){
    AsyncWebRewrite *r = _rewrites;
    _rewrites = r->next;
    delete r;
  }
  _rewrites = NULL;
  
  while(_handlers != NULL){
    AsyncWebHandler *h = _handlers;
    _handlers = h->next;
    delete h;
  }
  _handlers = NULL;
  
  if (_catchAllHandler != NULL){
    handler->onRequest(NULL);
    handler->onUpload(NULL);
    handler->onBody(NULL);
  }
}

AsyncWebServer::~AsyncWebServer(){
  reset();
  delete _catchAllHandler;
}
```

If you want to remove handler maybe another function should be added
```cpp
bool AsyncWebServer::removeHandler(AsyncWebHandler *handler){
  if(handler == _handlers){
    _handlers = _handlers->next;
    return true;
  }
  AsyncWebHandler *h = _handlers;
  while(h != NULL){
    if(handler == h->next){
      AsyncWebHandler *d = h->next;
      h->next = d->next;
      delete d;
      return true;
    }
  }
  return false;
}
```
 Same for rewrites I guess

me-no-dev:
Not all handlers are AsyncCallbackWebHandler :)

me-no-dev:
BTW to reset one of the _catchAllHandler callbacks, you can simply server.onUpload(NULL) to not handle it anymore

src/WebServer.cpp
```diff
@@ -66,6 +55,23 @@ AsyncWebRewrite& AsyncWebServer::addRewrite(AsyncWebRewrite* rewrite){
...
+    return true;
+  }
+  AsyncWebRewrite *r = _rewrites;
+  while(r != NULL){
+    if(rewrite == r->next){
+      AsyncWebRewrite *d = r->next;
+      r->next = d->next;
+      delete d;
+      return true;
+    }
```
me-no-dev: need to add ```r = r->next``` or while will loop forever
same for the method below

README.md
```diff
@@ -807,6 +807,33 @@ server.on("/scan", HTTP_GET, [](AsyncWebServerRequest *request){
 });
 ```
 
+### Remove handlers and writers
+
```
me-no-dev: Those are "rewrites" and not "writers". Please edit the README.

me-no-dev:
Thanks :)

PR #95, opened 2016-11-18
Template container instead of direct pointer manipulation. 

nut-code-monkey:
Instead of StringArray and direct pointer manipulation added template LinkedList container.
It's allow for( : ) enumeration and basic add/remove functionality. 
Removed duplicated code.
'OnRemove' block allow handle items that will be removed and for example delete pointers in single place.
Const reference to String as method parameters to prevent additional copy and memory allocation when String passed by value


me-no-dev:
wow! quite the pull request :D 
Don't you just love it when Arduino bites you with it's nonsense :D 
Restructuring includes and forward declarations usually work ;) If you can get away with including StringArray.h in the headers, things should go better

nut-code-monkey:
Fixed issue with travis build by removing c++ std:: includes. It's odd, because on mac adding std:: includes not produce any build issues for ESP8266.
Anyway is still enough to quick enumeration using for( : ).

me-no-dev:
you have tested everything correct?
do I need to test anything here? I really have no time to spare for ESP8266 at the moment

nut-code-monkey:
It's working well in my project. Yes, pool request looks very massy. But most changes - is 'const' and 'reference' parameter types.
Template container works as expected. I'm still not sure should i use __attribute__((always_inline)) for template methods or not. I't must prevent additional method call but will increase binary size. So for now it's not used.

me-no-dev:
pff I totally forgot about this PR, sorry :)

PR #107, opened 2016-12-05
Adding example of global & class functions for request handler

hreintke:


README.md
```diff
@@ -1020,6 +1021,57 @@ void loop(){
 }
 ```
 
+### Setup request handlers
```
me-no-dev: Title here should be the same as the one in the table of contents :)

PR #120, opened 2017-01-22
Convert resursive call to iterations, reduce stack usage

Adam5Wu:
I noticed, while trying to investigate crash stack trace (BTW, thanks so much for the exception decoder!), that the stack contains a great number (10+) of recursive calls to _onData().

While this did not cause crashes, I found it is quite simple to convert the parsing logic to iterative and thereby saving some stack spaces. :D


me-no-dev:
Great :)

PR #129, opened 2017-02-16
Events Source Message Queue

sticilface:
- Fixes problem where events sent close together are silently dropped.

### DIFF
--- a/src/AsyncEventSource.cpp
+++ b/src/AsyncEventSource.cpp
@@ -129,7 +129,7 @@ AsyncEventSourceMessage::~AsyncEventSourceMessage() {
 }
 
 void AsyncEventSourceMessage::ack(size_t len, uint32_t time) {
-  if(_len == len && _sent = _len){
+  if(_len == len && _sent == _len){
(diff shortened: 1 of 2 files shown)


sticilface:
This is based on the WS implementation.  Although I've removed the virtualisation of the basic message.  

There is no timeout removal of messages from queue.  yet. 

One thing I was considering was to maybe make the messages independent of the AsyncEventSourceClient.    In events, when a message is created all the clients get attached to it.  When ACKed they remove themselves from the message, when none are left, the message can be deleted.   I was also thinking about this strategy for web sockets as well.   The reason being that at the moment there is a malloc call per message for every client.  This is an unnecessary duplication of the data and may be significant if there is a lot.  If clients register with a message, then each message is only created once. 

Does that make sense?

sticilface:
Adding 100 messages one after the other works just fine.

me-no-dev:
are the two PRs overlapping?

sticilface:
i just realised that.  darn!  hate git sometimes

PR #130, opened 2017-02-16
Allow creation of WS messages with direct access to underlying buffer…

sticilface:
Prevents duplication of memory when creating messages.

- Move AsyncWebSocketBasicMessage into header
- Constructor without buffer and len `AsyncWebSocketBasicMessage(uint8_t opcode, bool mask)`
- methods to access `buffer()`, and `reserve()` buffer.

example using ArduinoJson

```cpp
    DynamicJsonBuffer jsonBuffer;
    JsonObject& root = jsonBuffer.createObject();
    AddJsonItems(root);
    size_t len = root.measureLength();

    AsyncWebSocketBasicMessage * message = new AsyncWebSocketBasicMessage();

    if (message) {
        if (message->reserve(len)) {
            root.printTo((char *)message->buffer(), len + 1); //  Required to print last char. 
            client->message(message);
        }
    }
```

### DIFF
--- a/src/AsyncWebSocket.cpp
+++ b/src/AsyncWebSocket.cpp
@@ -167,66 +167,81 @@ class AsyncWebSocketControl {
  * Basic Buffered Message
  */
 
-class AsyncWebSocketBasicMessage: public AsyncWebSocketMessage {
-  private:
-    uint8_t * _data;
-    size_t _len;
-    size_t _sent;
-    size_t _ack;
-    size_t _acked;
 
-  public:
-    AsyncWebSocketBasicMessage(const char * data, size_t len, uint8_t opcode=WS_TEXT, bool mask=false)
-      :_len(len)
-      ,_sent(0)
-      ,_ack(0)
-      ,_acked(0)
-    {
-      _opcode = opcode & 0x07;
-      _mask = mask;
-      _data = (uint8_t*)malloc(_len+1);
-      if(_data == NULL){
-        _len = 0;
-        _status = WS_MSG_ERROR;
-      } else {
-        _status = WS_MSG_SENDING;
-        memcpy(_data, data, _len);
-        _data[_len] = 0;
-      }
+AsyncWebSocketBasicMessage::AsyncWebSocketBasicMessage(const char * data, size_t len, uint8_t opcode, bool mask)
+  :_len(len)
+  ,_sent(0)
+  ,_ack(0)
+  ,_acked(0)
+{
+  _opcode = opcode & 0x07;
+  _mask = mask;
+  _data = (uint8_t*)malloc(_len+1);
+  if(_data == NULL){
+    _len = 0;
+    _status = WS_MSG_ERROR;
+  } else {
+    _status = WS_MSG_SENDING;
+    memcpy(_data, data, _len);
+    _data[_len] = 0;
+  }
+}
+AsyncWebSocketBasicMessage::AsyncWebSocketBasicMessage(uint8_t opcode, bool mask)
+  :_len(0)
+  ,_sent(0)
+  ,_ack(0)
+  ,_acked(0)
+{
+  _opcode = opcode & 0x07;
+  _mask = mask;
+  
+}
+
+AsyncWebSocketBasicMessage::~AsyncWebSocketBasicMessage() {
+  if(_data != NULL)
+    free(_data);
+}
+
+ void AsyncWebSocketBasicMessage::ack(size_t len, uint32_t time)  {
+  _acked += len;
+  if(_sent == _len && _acked == _ack){
+    _status = WS_MSG_SENT;
+  }
+}
+ size_t AsyncWebSocketBasicMessage::send(AsyncClient *client)  {
+  if(_status != WS_MSG_SENDING)
+    return 0;
+  if(_acked < _ack){
+    return 0;
+  }
+  if(_sent == _len){
+    if(_acked == _ack)
+      _status = WS_MSG_SENT;
+    return 0;
+  }
+  size_t window = webSocketSendFrameWindow(client);
+  size_t toSend = _len - _sent;
+  if(window < toSend) toSend = window;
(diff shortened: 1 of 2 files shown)


sticilface:
example of sending a binary message, placing a circular buffer into the right order, without having to create a temporary buffer beforehand.  

```cpp
void send_temphx(AsyncWebSocketClient * client)
{
    AsyncWebSocketBasicMessage * message = new AsyncWebSocketBasicMessage(WS_BINARY);

    if (message) {

        if (message->reserve(temphx.items * sizeof(float))) {

            uint8_t * data = message->buffer();

            for (uint16_t i = 0 ; i < temphx.items; i++) {

                uint16_t position = (temphx.out + i ) % PID_Vars.graphsize;
                uint8_t * array = reinterpret_cast<uint8_t*>(&temphx.temps[position]);

                for (uint8_t i = 0 ; i < sizeof(float) ; i++) {
                    data[position * sizeof(float) + i] = array[i];
                }
            }

            if (client) {
                client->message(message);
            } else {
                ws.messageAll(message);
            }
        }
    }
}
```

sticilface:
In principle this works, however there is a bug as there are several assumptions made. 
1) All messages are created using new.  This is how they are done in the underlying websockets lib, so that on ACK they are deleted.   
2) messages are not added if the client is not connected.
3) who deletes the message if it is not added to a client?  the caller, or the websocket lib. 

4) Currently this function below does not really work, as the ownership of the message is not clear.  There is no copy constructor that can copy the _data, and allocate the memory for the new message.  Ideally a message would be created, and then copied to all the available and connected clients, with each client deleting its own message.     

```
void AsyncWebSocket::messageAll(AsyncWebSocketMessage *message){
  for(const auto& c: _clients){
    if(c->status() == WS_CONNECTED)
      c->message(message);
  }
}
```

So what is the best way to fix this?

Provide copy constructors so messages can be copied.  then the above function would work. However, multiple copies of the same message is not efficient.  I propose abstracting out messages to be independent of clients, and client subscribe to a message, removing themselves on ACK.  When there are no subscribed clients / after a timeout the message is deleted. 

Provide access to the client iterators so the user can go though them and create their own messageAll.   

any thoughts?

hreintke:
@sticilface : 
Not knowing detail on the PR but would using a shared pointer to the message solve a number of the issues above ?

sticilface:
You would think so.  However, I've had 2 separate nightmares with using shared_ptr and memory leaking, and I could not for the life of me figure out what was causing it!  

I also do not think that this strategy I've pushed above will work.  I have implemented it but it gets very complicated very quickly with 1) creating messages outside of the main WS API and then if they are not added to a WS client, no one is managing it deletion (and there is no way of knowing if it has been added or not 2) the lifetime of the message is dependent on the ACK, and this may come at any point. 

My solution that I have implemented and is working, is to separate out the msg buffer as its own class, whose lifetime is managed by the WS server which can perform cleanup of unused buffers.  I've created a separate msg class, that takes this buffer in its constructor

```cpp
class AsyncWebSocketMessageBuffer {
  private:
    uint8_t * _data;
    size_t _len;
    bool _lock; 
    uint32_t _count;  

  public:
    AsyncWebSocketMessageBuffer();
    AsyncWebSocketMessageBuffer(size_t size);
    AsyncWebSocketMessageBuffer(uint8_t * data, size_t size); 
    AsyncWebSocketMessageBuffer(const AsyncWebSocketMessageBuffer &); 
    ~AsyncWebSocketMessageBuffer(); 
    void operator ++(int i) { _count++; }
    void operator --(int i) {  if (_count > 0) { _count--; } ;  }
    bool reserve(size_t size);
    void lock() { _lock = true; }
    void unlock() { _lock = false; }
    uint8_t * get() { return _data; }
    size_t length() { return _len; }
    uint32_t count() { return _count; }
    bool canDelete() { return (!_count && !_lock); } 
    friend AsyncWebSocket; 

};
``` 

and the msg class
```cpp
class AsyncWebSocketMultiMessage: public AsyncWebSocketMessage {
  private:
    uint8_t * _data;
    size_t _len;
    size_t _sent;
    size_t _ack;
    size_t _acked;
    AsyncWebSocketMessageBuffer * _WSbuffer; 
public:
    AsyncWebSocketMultiMessage(AsyncWebSocketMessageBuffer * buffer, uint8_t opcode=WS_TEXT, bool mask=false); 
    virtual ~AsyncWebSocketMultiMessage() override;
    virtual bool betweenFrames() const override { return _acked == _ack; }
    virtual void ack(size_t len, uint32_t time) override ;
    virtual size_t send(AsyncClient *client) override ;
    // uint8_t * buffer() { return _data; }
    // bool reserve(size_t size);
};
```

to get a buffer you request it through the server so that it can manage its lifetime

```cpp

    //  messagebuffer functions/objects. 
    AsyncWebSocketMessageBuffer * makeBuffer(size_t size = 0); 
    AsyncWebSocketMessageBuffer * makeBuffer(uint8_t * data, size_t size); 
    LinkedList<AsyncWebSocketMessageBuffer *> _buffers;
    void _cleanBuffers(); 
```

This set up allows shared buffers between messages sent using *All commands, but prevents leaks when a message is not actually added to a client, whilst still allowing direct manipulation of the underlying buffer.   example of new textAll

```cpp
void AsyncWebSocket::textAll(const char * message, size_t len){
  AsyncWebSocketMessageBuffer * WSBuffer = makeBuffer((uint8_t *)message, len); 
  for(const auto& c: _clients){
    if(c->status() == WS_CONNECTED)
      c->text(WSBuffer);
  }
  _cleanBuffers();  //  cleans up if the message is not added to any clients. 
}
```

Still to consider: 
- refactoring all the printf and all commands to print directly to this buffer
- abstract further to allow web events messaging system to use the same methods

sticilface:
ok here is my working version take a look @me-no-dev see what you think.  Different approach, but it is all working. 
https://github.com/sticilface/ESPAsyncWebServer/tree/wsbuffer

PR #134, opened 2017-02-21
  flash string malloc memory leak

sticilface:
malloc is not freed when using web sockets F().

Red line demonstrate time of WS client connection.  Sketch sends large F() string every second. 

Free heap Before fix
screen shot 2017-02-21 at 10 22 58
Free heap After fix
screen shot 2017-02-21 at 10 26 38


sticilface:
oh crap ignore this pull.  its included all my other fixes as well.  Thought i'd excluded them.

PR #139, opened 2017-02-23
Some enhancements

wkraft-fablabka:
- Added "SAVE-Button
- Upgraded to ACE 1.2.6
- Added additional field, which shows name of the current edited File
- Reset upload path after uploading or creating a File (avoid confusion about which file is shown in editor)
- added the possibility, to run the editor completly from flash. When ace.js.gz or ace.js and the corresponding files are found on SPIFFS, the editor can be used in AP mode also.  Otherwise the files are searched on cloudflare like before.
- added all gzipped js-files in data directory
- added an exclude list, containing wildcards, telling which files should not be listed in the editor (to prevent deletion of needed files)
- minified the embedded HTML/Javascript with https://kangax.github.io/html-minifier/ before gzipping.  With all additions the resulting hex-code is now 700 bytes smaller
- embedded the plain html as documentation


src/SPIFFSEditor.cpp
```diff
@@ -435,3 +393,46 @@ void SPIFFSEditor::handleUpload(AsyncWebServerRequest *request, const String& fi
...
+      continue;
+    }
+    return false;
+  }
+  while (*pattern=='*'){pattern++;}  
+  return (*pattern == 0);
+}
+
+
+bool isHidden( const char *filename, const char *excludeList) {
```
me-no-dev: why not ```isExcluded```?

hreintke:
Hi,

I have the same yield issues (and delay as well) as you mentioned in my project.
Submitted a issue/question on this on the esp8266/Arduino github.

https://github.com/esp8266/Arduino/issues/3042

me-no-dev:
OK I have finally found some time to go through this. Did some changes to be inline with the latest commits and created a new PR for it. Closing this one now :)

PR #144, opened 2017-03-05
Web Sockets memory improvements : shared message buffer, direct buffer access

sticilface:
…messages to multiple clients.

Add ability to create a WS buffer directly and write to it, saving duplication of message in RAM, then send it.  example below is an arduinoJson message.

```cpp
void sendDataWs(AsyncWebSocketClient * client)
{
    DynamicJsonBuffer jsonBuffer;
    JsonObject& root = jsonBuffer.createObject();
    root["a"] = "abc";
    root["b"] = "abcd";
    root["c"] = "abcde";
    root["d"] = "abcdef";
    root["e"] = "abcdefg";
    size_t len = root.measureLength();
    AsyncWebSocketMessageBuffer * buffer = ws.makeBuffer(len); //  creates a buffer (len + 1) for you.
    if (buffer) {
        root.printTo((char *)buffer->get(), len + 1);
        if (client) {
            client->text(buffer);
        } else {
            ws.textAll(buffer);
        }
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -790,6 +790,32 @@ const uint8_t flash_binary[] PROGMEM = { 0x01, 0x02, 0x03, 0x04 };
 client->binary(flash_binary, 4);
 ```
 
+### Direct access to web socket message buffer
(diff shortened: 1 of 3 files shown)


me-no-dev:
@sticilface please do add the example to README ;)

PR #149, opened 2017-03-24
The previous server event need not be ACKed to send the next one (it will be queued)

me21:
I've noticed that ESPAsyncWebServer does not send next server event until the previous one is ACKed. When firmware sends server events too frequently, it could cause memory exhaustion (for example, Windows ACKs TCP frame after 200 ms; if firmware generates more than 5 events per second, it will consume RAM of ESP8266).
After my changes, there can be more than one server event sent in single TCP frame.

### DIFF
--- a/src/AsyncEventSource.cpp
+++ b/src/AsyncEventSource.cpp
@@ -107,20 +107,15 @@ static String generateEventMessage(const char *message, const char *event, uint3
 // Message
 
 AsyncEventSourceMessage::AsyncEventSourceMessage(const char * data, size_t len)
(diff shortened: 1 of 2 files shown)


me-no-dev:
OK, so you say this sends the queued events in one packet (if possible) when the ack of the last packet comes?

me21:
Yes, at least it did when used in my firmware.

me21:
I've fixed a small bug in AsyncEventSourceMessage::ack function when the peer ACKed the message only partially (it did not increment _acked counter).

PR #159, opened 2017-04-20
Fixing compiler warnings exposed with -Wall and -Wextra

trygvis:
AsyncWebHandler::setAuthentication: return is required.
AsyncWebSocketMultiMessage::~AsyncWebSocketMultiMessage: error: enumeral and non-enumeral type in conditional expression.


me-no-dev:
thanks :)

PR #173, opened 2017-05-22
Added onDisconnect

faceless2:
I've added an "onDisconnect" handler to the AsyncCallbackWebHandler class. The reasoning behind this is when uploading content as part of the request body, resources may need to be allocated to handle this. But if the upload terminates early there was no way to free those resources (well, none that I could idenitfy).

With this patch I can now do this:

```
webserver.on("/write", HTTP_POST, [](AsyncWebServerRequest *request) {
        // handle request
    }, NULL, [](AsyncWebServerRequest *request, uint8_t *data, size_t len, size_t index, size_t total) {
        // allocate resources
    }).onDisconnect([](AsyncWebServerRequest *request) {
        // free resources
    });
```

and know that the "onDisconnect" method will always be called, even on error.

### DIFF
--- a/src/ESPAsyncWebServer.h
+++ b/src/ESPAsyncWebServer.h
@@ -315,6 +315,7 @@ class AsyncWebHandler {
     virtual void handleRequest(AsyncWebServerRequest *request __attribute__((unused))){}
     virtual void handleUpload(AsyncWebServerRequest *request  __attribute__((unused)), const String& filename __attribute__((unused)), size_t index __attribute__((unused)), uint8_t *data __attribute__((unused)), size_t len __attribute__((unused)), bool final  __attribute__((unused))){}
     virtual void handleBody(AsyncWebServerRequest *request __attribute__((unused)), uint8_t *data __attribute__((unused)), size_t len __attribute__((unused)), size_t index __attribute__((unused)), size_t total __attribute__((unused))){}
+    virtual void handleDisconnect(AsyncWebServerRequest *request __attribute__((unused))){}
 };
 
 /*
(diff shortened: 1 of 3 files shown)


me-no-dev:
maybe implement this in the base web handler class?

me-no-dev:
Actually wait... you should be allocating those resources in the Response classes and not in the Handler classes. Basically you will get the disconnect while you are responding, request disconnects are handled already.

luc-github:
@faceless2 @me-no-dev 
I also need such feature to know when a request is closed for any reason (succeed or error).
On Async I think such mecanism is mandatory and we need a solution, so here my proposal

Instead of adding a new parameter in request creation, the call is defined in request itself to make it as flexible as possible:

in ESPAsyncWebServer.h
```
class AsyncWebServerRequest {
    private:
    std::function<void(void)> _onDisconnectfn;
    ....
    public:
    void onDisconnect (std::function<void(void)> fn);
    ....
  ````  
  in WebRequest.cpp
```
    void AsyncWebServerRequest::onDisconnect (std::function<void(void)> fn){
    _onDisconnectfn=fn;
    }
```
```
void AsyncWebServerRequest::_onDisconnect(){
  //os_printf("d\n");
  if(_onDisconnectfn) {
      _onDisconnectfn();
    }
  _server->_handleDisconnect(this);
}
```

So in any handler that need a disconnect the customized function can be added or not:
```
request->onDisconnect([](){Serial.println("We are done!");});
```

What do you think ? I know the style little bit raw, any chance to get such thing implemented ?

PR #176, opened 2017-06-09
Added simple template processor to AsyncFileResponse. Optimized chunked responses.

me21:
Unzipped files may have template placeholders like `%TEMPLATE_VAR%` inside. If callback is specified in `AsyncFileResponse` constructor call, it will be used to replace these with actual strings.
The prototype of callback is `String(const String&)`, i.e. it gets variable name and returns its value.
Template variables' delimiter is currently percent sign (`%`).
Maximal placeholder length is 32 chars (chosen somewhat arbitrarily, it may be stored on stack during processing). It is not guaranteed that placeholders longer than that will be processed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ To use this library you might need to have the latest git versions of [ESP8266](
 		- [Respond with content coming from a Stream and extra headers](#respond-with-content-coming-from-a-stream-and-extra-headers)
 		- [Respond with content coming from a File](#respond-with-content-coming-from-a-file)
 		- [Respond with content coming from a File and extra headers](#respond-with-content-coming-from-a-file-and-extra-headers)
+		- [Respond with content coming from a File containing templates](#respond-with-content-coming-from-a-file-containing-templates)
 		- [Respond with content using a callback](#respond-with-content-using-a-callback)
 		- [Respond with content using a callback and extra headers](#respond-with-content-using-a-callback-and-extra-headers)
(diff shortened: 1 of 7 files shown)


me21:
Hmm, I'm not sure what's gone wrong, I've checked that both library and example is built successfully on my machine before uploading.

me21:
There. Because <Arduino.h> conflicts with <vector>, it causes build errors when <vector> is included after <Arduino.h>.

Since my commit makes use of vector internally, it is now necessary to include <ESPAsyncWebServer.h> before <Arduino.h>. Not good, obviously. Can it be somehow resolved without ditching vector class?

me21:
Done :)

me21:
Perhaps it should be done as another subclass of AsyncWebServer response? Or not?

me-no-dev:
This is great! and people have been asking for it! But I think this should be done in a way that not only files can be templates. Why not store the template in progmem or something. And also could you please separate the chunked encoding enhancement and PR it?

me21:
I will do it, I just don't have enough free time right now.
Just thought that using templates in files is the most common use case. Of course, the approach can be expanded to progmem etc.

me21:
I'm almost done with it, BTW. I just haven't updated documentation yet. Will open another pull request and close this one when I'm done.

me21:
Done, template processing is now in #189, and optimized chunked responses are in #188.

PR #180, opened 2017-06-20
Cache-Control header value needs = separator (not : )

jej:
As specified here https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9 the separator for Cache-Control header values is `=`


me-no-dev:
Thanks!

PR #183, opened 2017-06-22
New project based on AsyncWebServer

omersiar:


src/WebAuthentication.cpp
```diff
@@ -205,9 +205,9 @@ bool checkDigestAuthentication(const char * header, const char * method, const c
     }
   } while(nextBreak > 0);
 
-  String ha1 = (passwordIsHash) ? String(password) : myUsername + ":" + myRealm + ":" + String(password);
+  String ha1 = (passwordIsHash) ? String(password) : stringMD5(myUsername + ":" + myRealm + ":" + String(password));
```
me-no-dev: this is already merged. how come it shows up again?

omersiar: I really do not know, currently i'm on windows machine and pushed them via Githup App. I try to rebase. Hang on.

PR #185, opened 2017-06-23
Fix url decoding in POST / GET requests

mrk-its:
Currently parsing GET / POST parameters containing '&' and '=' characters (in parameter name or value) is broken (full url is decoded instead of url parts). This solution is similar to https://github.com/me-no-dev/ESPAsyncWebServer/pull/84 but more complex - it allows for '&=' characters in parameter name and fixes parsing of POST payload.


me-no-dev:
thanks :)

PR #189, opened 2017-07-19
Added simple template processor to response classes.

me21:
Unzipped files in SPIFFS, Streams, PROGMEM strings, callback/chunked responses may have template placeholders like %TEMPLATE_VAR% inside. If callback is specified in ```AsyncAbstractResponse``` children constructor call, it will be used to replace these with actual strings. For library users, there are corresponding arguments in ```AsyncWebServerRequest::send``` and ```AsyncWebServerRequest::beginResponse``` method families. 
The prototype of callback is ```String(const String&)```, i.e. it gets variable name and returns its value.
Template variables' delimiter is currently percent sign (```%```).
Maximal placeholder length is 32 chars (chosen somewhat arbitrarily, it may be stored on stack during processing). It is not guaranteed that placeholders longer than that will be processed.


me-no-dev:
Can you please fix the conflict so I can merge :)

UlliBien:
still doesn't work:
`  request->send(SPIFFS, "/t.svg", String(), false, processor);
`
and 
```
String processor(const String& var) {
  Serial.println("************* processor");
  Serial.println(var);
  return var;
}
```
gives 

> ************* processor
üNOOü]]]ûNOOúYYYý^^^ÿ^^^ÿ^^^ÿ^^^
************* processor
ûNNNù]]]úZZZü^^^ÿ^^^ÿ^^^ÿ^^^ÿ^^^
************* processor
üYYYü^^^ÿ^^^ÿ^^^ÿ^^^ÿYYYþ$&&ü<ESC><GS><GS>

> 
It seems that 'var' is not filled.

Antother problem that makes the solution useless is that method `processor` is called more than once. I think it might be because of splitting the file in 3 chunks. You cannot detect a template that is splitted!

me21:
Okay, can you please prepare minimal example that shows the bug? What is the contents of your file t.svg? Is it a text file or SVG image?
Actually, it should detect split templates, it has a read-ahead cache for that...

UlliBien:
t.svg is a text file containing xml (svg) code. Iadded it to the comment.
[t.zip](https://github.com/me-no-dev/ESPAsyncWebServer/files/1270550/t.zip)
I downloaded the library again to get the newest version and constructed a small project. 
[AsyncWebTest.zip](https://github.com/me-no-dev/ESPAsyncWebServer/files/1270696/AsyncWebTest.zip)
Without special handlings for t.svg, i get the expected result: 
good
To handle t.svg I added in septup() `  server.on("/t.svg", HTTP_ANY, handle_T_Request);`
and outside any container:
```
String T_Processor(const String& var) {
  Serial.println("************* processor");
  Serial.println(var);
  return var;
}
void handle_T_Request(AsyncWebServerRequest * request) {
  //Send index.htm with template processor function
  Serial.println("In handle_T_Request");
  request->send(SPIFFS, "/t.svg", String(), false, T_Processor);
}
```
Now a problem appears. T_Processor is not called! But the file content is repeated incomplete several times. In file BrowsweInside.txt 
[BrowserInside.txt](https://github.com/me-no-dev/ESPAsyncWebServer/files/1270729/BrowserInside.txt)
 I copied the website source code out of FireFox. Here a screenshot: 
bad

UlliBien:
I have a question to "Actually, it should detect split templates, it has a read-ahead cache for that..."
How do I have to mark a template? Is it %template%, like shown in the examples? How can I mask "%", if a file contains "%" as a regular symbol?

me21:
Yes, it's %template%, like in examples. To mask percent sign, repeat it twice: %%.

I'll take a look on your example on Monday, thanks!

UlliBien:
You should add the rules of template handling to the docomentation.

me21:
@UlliBien it is already done, see https://github.com/me-no-dev/ESPAsyncWebServer#template-processing

Meanwhile, the issue has been fixed in #219. Try yourself.

UlliBien:
Works perfectly now!
THX!

PR #193, opened 2017-07-24
When max number of SPIFFS open files is exceeded: return 404...

me21:
even when onNotFound callback is set and let browser re-request the URI (at least Firefox does so).

Possibly 429 Too Many Requests should be returned?..

### DIFF
--- a/src/WebHandlers.cpp
+++ b/src/WebHandlers.cpp
@@ -133,24 +133,25 @@ bool AsyncStaticWebHandler::_fileExists(AsyncWebServerRequest *request, const St
   String gzip = path + ".gz";
 
   if (_gzipFirst) {
-    request->_tempFile = _fs.open(gzip, "r");
-    gzipFound = request->_tempFile == true;
+    gzipFound = _fs.exists(gzip);
     if (!gzipFound){
-      request->_tempFile = _fs.open(path, "r");
-      fileFound = request->_tempFile == true;
+      fileFound = _fs.exists(path);
     }
   } else {
-    request->_tempFile = _fs.open(path, "r");
-    fileFound = request->_tempFile == true;
+    fileFound = _fs.exists(path);
     if (!fileFound){
-      request->_tempFile = _fs.open(gzip, "r");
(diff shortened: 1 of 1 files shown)


me21:
I've noticed that before my patch, when onNotFound callback is not set, ESPAsyncWebServer returns status code 500.

To sum up, here's behaviour table when there are too many open files on SPIFFS:

onNotFound set? | Before patch | After patch
-------------|-------------|-------------
Yes | called onNotFound | 404
No | 500 | 404

Perhaps it should be changed back to 500? Or even 429?..

me21:
Another solution would be to keep files open only when next data chunk needs to be sent, and to close them in between, but I have no idea how big the performance hit is going to be.

me-no-dev:
overall the reason why I do not use fs.exists() is because calling that and then opening takes a lot of time, especially on a large SPIFFS partition. I'd recommend to keep the old code ;)
Also in your change, you do not check if the file was actually opened

me21:
Well, I guess I can look at request URL in onNotFound callback and then check fs.exists() myself. I'm OK with keeping the old code here.

By the way, there is another check if file has been succesfully opened: in AsyncStaticWebHandler::handleRequest, so checking the same in AsyncStaticWebHandler::_fileExists is not necessary.

me21:
I think it's worth documenting that onNotFound callback will be called if max number of open files is exceeded even when the file exists. At least it was not obvious for me.

Or maybe there's another way to solve that problem. I'll update this pull request if I come up with something.

ccoenen:
Choosing `429` or `500` is pretty easy:

`429` is a client error - you'd use it if someone were DoS'ing your service or went above a defined API-Request-Limit.
the `500` class is a server error - I believe this is more appropriate here. The server (i.e. the ESP) can't serve the request, but it's not _really_ the fault of the client. One could use `500`, which does not really imply a reason or `503` which would be "Service unavailable", which would mostly be true.

I also don't think the onNotFound callback is the right choice, it will just mask an error, making it harder for any user to pin down the problem.

PR #205, opened 2017-08-14
Add basic ESP32 support

vincasmiliunas:
Hi, I've patched up the code to get it running with the current Arduino-ESP32. It works both on ESP8266 and ESP32. It actually forces code paths only for ESP8266 or ESP32, is ESP31B any relevant?

You need patched ESPAsyncTCP to run this - https://github.com/vincasmiliunas/ESPAsyncTCP

At this time Arduino-ESP32 does provide any SPIFFS library [1], thus you need to download one from third party [2]. HAS_ARDUINO_ESP32_SPIFFS is used mark that it's available.
[1] - https://github.com/espressif/arduino-esp32/tree/master/libraries
[2] - https://github.com/copercini/arduino-esp32-SPIFFS

This patch uses new APIs or applies workarounds to fix the incompatibilies between ESP8266 and ESP32, like Arduino-ESP32 [1] destroying method names with C preprocesor macros, unlike Arduino-ESP8266 [2].
[1] - https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/pgmspace.h#L65
[2] - https://github.com/esp8266/Arduino/blob/master/cores/esp8266/pgmspace.h#L77

### DIFF
--- a/examples/ESP_AsyncFSBrowser/ESP_AsyncFSBrowser.ino
+++ b/examples/ESP_AsyncFSBrowser/ESP_AsyncFSBrowser.ino
@@ -98,7 +98,7 @@ void setup(){
   WiFi.mode(WIFI_AP_STA);
   WiFi.softAP(hostName);
   WiFi.begin(ssid, password);
-  if (WiFi.waitForConnectResult() != WL_CONNECTED) {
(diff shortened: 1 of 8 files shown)


vincasmiliunas:
me-no-dev has updated ESPAsyncWebServer with ESP32 support!

PR #208, opened 2017-08-18
Enhancements by @wkraft-fablabka

me-no-dev:
- Added "SAVE-Button
- Upgraded to ACE 1.2.6
- Added additional field, which shows name of the current edited File
- Reset upload path after uploading or creating a File (avoid confusion about which file is shown in editor)
- added the possibility, to run the editor completly from flash. When ace.js.gz or ace.js and the corresponding files are found on SPIFFS, the editor can be used in AP mode also.  Otherwise the files are searched on cloudflare like before.
- added all gzipped js-files in data directory
- added an exclude list, containing wildcards, telling which files should not be listed in the editor (to prevent deletion of needed files)
- minified the embedded HTML/Javascript with https://kangax.github.io/html-minifier/ before gzipping.  With all additions the resulting hex-code is now 700 bytes smaller
- embedded the plain html as documentation


hallard:
"added the possibility, to run the editor completly from flash. When ace.js.gz or ace.js and the corresponding files are found on SPIFFS, the editor can be used in AP mode also. Otherwise the files are searched on cloudflare like before"

This is a great feature, but I think, it should be done the reverse way, IE if ace source found on Internet (client  has internet connection) download from Internet, if ace source not found (mainly client connected on ESP AP) then load from flash, because the way it's implemented today if files are on flash all is loaded from flash.

may be 
```html
<script id='ace' src="/ace.js" type="text/javascript" charset="utf-8"></script>
<script>
  if  (typeof ace.edit == "undefined") {
    var script = document.createElement('script');
    script.src = "https://cdnjs.cloudflare.com/ajax/libs/ace/1.2.6/ace.js";
    script.async = false;
    document.head.appendChild(script);
  }
</script>
```
should be
```html
<script id='ace' src="https://cdnjs.cloudflare.com/ajax/libs/ace/1.2.6/ace.js" type="text/javascript" charset="utf-8"></script>
<script>
  if  (typeof ace.edit == "undefined") {
    var script = document.createElement('script');
    script.src = "/ace.js";
    script.async = false;
    document.head.appendChild(script);
  }
</script>
```
I let you decide what's the best implementation, but anyway, it's a really nice improvement  ;-)

PR #209, opened 2017-08-27
Address compilation warning.

arantius:
Specifically: "warning: comparison between signed and unsigned integer
expressions".  Make `i` the same type as `len`, which it is frequently
compared to.


me-no-dev:
thanks :)

PR #213, opened 2017-08-31
Something got lost during pull request #189. Fixes #211, updates documentation.

### LINKED ISSUE #211: "Respond with content coming from a File containing templates" not working
I tried the example for "**Respond with content coming from a File containing templates**". The callback method "**processor(..)**" is never called.

`request->send(SPIFFS, "/index.htm", String(), false, processor);` initializes a new intance of `AsyncAbstractResponse` class (`AsyncFileResponse` is derived from `AsyncAbstractResponse`). Field `_chunked` is set to true by the constructor. Via some other methods `AsyncAbstractResponse::_ack(..)` is called.

The problem is that `_ack(..)` does not call method `AsyncAbstractResponse::_fillBufferAndProcessTemplates(..) `which calls `_callback(..)` if `_chunked` is set to true. But it should! It calls `_fillBuffer(..)` instead. :-(

me21:


me-no-dev:
got it :)

PR #218, opened 2017-09-05
Fixed chunked file response length.

me21:


me21:
Hmm, it keeps on adding old commits which were already merged.

Issue #51, opened 2016-07-03
Enhancement: addHeader()


andig:
Please allow `addHeader` similar and generic to `setLastModified` and `setCache`\- for static handlers and EventSource. This will e.g. allow to specify the CORS headers for Ajax.

PR #230, opened 2017-09-17
Allow "default" headers to be defined for all responses.

tetious:
This PR adds a (global) singleton DefaultHeaders to allow adding headers that will automatically be included with all responses. 

I've tested this with EventSource and dynamic responses, but not yet with the Static response types, though it should work with anything that inherits from AsyncWebServerResponse, and I think that's all the Responses. :)

This should implement #51.


me-no-dev:
Nice! Yes it will work with any response and they all inherit AsyncWebServerResponse :) merging...

Issue #246, opened 2017-10-04
unnecessary last char in WS printfAll 


qc24:
After update library to actual version i found out that websocket printfAll doesn't work properly. Function adding unnecessary char on the end of sending string. I think it is issue in /src/AsyncWebSocket.cpp in printfAll because textAll work great.

PR #247, opened 2017-10-04
Update AsyncWebSocket.cpp

qc24:
It's resolving issue #246 but i'm not sure it's ok.


qc24:
I understand that makebuffer do len+1 and seted it to 0 before (in constructor), and if you add "makeBuffer(len+1)"  you get len+2 instead len+1, so line "AsyncWebSocketMessageBuffer * buffer = makeBuffer(len);" resolving issue

me-no-dev:
ahh you are right... sorry for delaying this

PR #266, opened 2017-10-25
Disabled request body parsing if the handler does nothing.

me21:
This will save memory and prevent crashes on large POST requests.


me21:
Hmm, I've noticed one drawback: if someone derives new subclasses of AsyncWebHandler which handle POST requests, he must not forget to redefine isRequestHandlerTrivial() function.

me-no-dev:
Nice :)

PR #143, opened 2017-03-04
Add basic auth for server static

tuanpmt:

PR #272, opened 2017-10-28
This re-enables previously removed WebSocket authentication which is …

omersiar:
…orginaly implemented PR #143


me-no-dev:
cool! :)

PR #280, opened 2017-11-04
Fix server method typo in README

UniquePassive:
This typo had me debugging for far too long.


me-no-dev:
:D thanks for digging it out!

PR #284, opened 2017-11-16
Allow callback when request disconnects

luc-github:


src/ESPAsyncWebServer.h
```diff
@@ -128,6 +128,7 @@ class AsyncWebServerRequest {
     AsyncWebHandler* _handler;
     AsyncWebServerResponse* _response;
     StringArray _interestingHeaders;
+    std::function<void(void)> _onDisconnectfn;
```
me-no-dev: why not typedef this like it is done in AsyncTCP?

luc-github: You means doing ?
`typedef std::function<void(void)> ArCallbackHandler;`

Because I was just lazy to add extra line ^_^ - but I can do it if you want :+1: 
Just confirm the name you want for the typedef :cat:

me-no-dev: maybe ArDisconnectHandler?

luc-github: Sure will do that -  thank you

luc-github:
I think it is ok now - let me know if anything

Issue #195, opened 2017-07-24
Document how to parse JSON POSTed to the ESP


probonopd:
What is the ESPAsyncWebServer equivalent of https://github.com/esp8266/Arduino/issues/1321#issuecomment-267676688
```
StaticJsonBuffer<200> jsonBuffer;
JsonObject& root = jsonBuffer.parseObject(server.arg("plain"));
```
?

The following _educated guess_, `jsonBuffer.parseObject(request->getParam("plain", true)->value());`, seems __not__ to work for me:

```
  server.on("/api/printer/command", HTTP_POST, [](AsyncWebServerRequest * request) {
    const size_t bufferSize = JSON_OBJECT_SIZE(1) + 250;
    DynamicJsonBuffer jsonBuffer(bufferSize);
    JsonObject& root = jsonBuffer.parseObject(request->getParam("plain", true)->value());
    const char* command = root["command"];
    request->send(204);
  });
```

Is this documented somewhere? If not, please document it.

jeremypoulter:
For what it is worth, I have got this to work, have a look [here](https://github.com/jeremypoulter/SmartPlug/blob/master/src/web_ui.cpp#L111:L151)

probonopd:
So you are saying it should be `JsonObject& root = jsonBuffer.parseObject(request->getParam("body", true)->value());`? Where do you have that information from?

jeremypoulter:
I think so (was a while ago when I last looked at this). IIRC I just cut and paste the content of the not found handler from the example. This outputs loads of useful info about the request.

probonopd:
Thank you. Can't seem to get it to work yet, though...

stelgenhof:
Tried it and works perfectly!

probonopd:
Apparently there is more than meets the eye. `request->getParam("body", true)->value()` does not work for me since apparently there is no `body` to begin with...

If I send the following request `curl -H "Content-Type: application/json" -X POST -d '{"command": "Hello"}' http://192.168.0.22/` to the ESP, which looks like this:

```
POST / HTTP/1.1
User-Agent: curl/7.37.0
Host: 127.0.0.1
Accept: */*
Content-Length: 20

{"command": "Hello"}
```

then the sketch returns `No body?!`:

```
#include <ESP8266WiFi.h>

#include <ESPAsyncTCP.h>
#include <ESPAsyncWebServer.h>
[...]

PR #287, opened 2017-11-17
Added handler for correct JSON deserialization, addresses #195

avonwyss:
While searching for a proper solution to handle POST JSON requests I have found different answers but none seemed to be right (some not working at all) and in line with the async approach. All of the ones I found do their processing in the `onRequestBody` handler (either on the server or as argument in the callback handler), but this approach is not a good one:

- If the JSON data does not come in all at once (e.g. in one call to `onRequestBody`), the code will terribly break.
- There is no way how the data being processed in the `onRequestBody` can be associated (and managed) across multiple calls, or to pass the data along to `handleRequest`. This makes it impossible to properly deal with the data.

My take is to implement a generic JSON handler `AsyncCallbackJsonWebHandler` which follows the same pattern as the `AsyncCallbackWebHandler`, but does not expose file and body handling as callbacks but rather handles them. The `handleRequest` function receives a `JsonVariant` reference without the need to manually parse anything. Invalid requests, or requests exceeding the allowed content length, are rejected gracefully.


me-no-dev:
There are request variables that you can use to store a pointer to wherever you want to collect the data (https://github.com/me-no-dev/ESPAsyncWebServer/blob/master/src/ESPAsyncWebServer.h#L189-L190) but this approach is good too :) Can you please add the example to README as well?

avonwyss:
@me-no-dev I added the example, and a notice regarding `_tempObject` too (I'd used that as storage pointer if I had found it). However, since the handler takes care of all the data handling and parsing I do think it still has its merits. ;)

me-no-dev:
I do not argue with that :) Just wanted to point out that there is a place to store a pointer for every request

README.md
```diff
@@ -284,6 +286,22 @@ void handleBody(AsyncWebServerRequest *request, uint8_t *data, size_t len, size_
...
+If needed, the `_tempObject` field on the request can be used to store a pointer to temporary data (e.g. from the body) associated with the request. If assigned, the pointer will automatically be freed along with the request.
+
+### JSON body handling with ArduinoJson
+Endpoints which consume JSON can use a special handler to get ready to use JSON data in the request callback:
+```cpp
+#include "AsyncJson.h"
+#include "ArduinoJson.h"
+
+AsyncCallbackJsonWebHandler* handler = new AsyncCallbackJsonWebHandler();
+handler->setUri("/rest/endpoint");
```
me-no-dev: I'm thinking that the URI should be a part of the AsyncCallbackJsonWebHandler constructor (almost everything else in the lib works that way)

me-no-dev:
sorry to bug you again :) but how about adding the ```onRequest``` as a second argument to the constructor? That would make it inline with rest of the lib. Sorry I did not notice that the first time :(

avonwyss:
No problem, done.

However, I do have to disagree about this being in line with the rest of the lib; I based the call style on the  `AsyncCallbackWebHandler` class: Only the `on(...)` method in the `AsyncWebServer` class does get the arguments directly, not the constructors. However, I have not added a `on` method overload to `AsyncWebServer` to avoid a dependency on `ArduinoJson`.

baggior:
@me-no-dev sorry, when are you intending to merge this code on the main branch?

me-no-dev:
Totally my fault. Have been overwhelmed by work and the fact that I moved to another country... merging as soon as travis is done testing

PR #310, opened 2018-01-08
fix makeBuffer example to avoid memory corruption

thunderace:
In Readme example, len bytes are allocated and len+1 bytes are written : memory corruption


thunderace:
my bad, no memory corruption here...

PR #311, opened 2018-01-09
Change JSON MIME type to application/json

merlinschumacher:
Hey,

the correct mimetype for JSON would be application/json not text/json


me-no-dev:
OMG it's been a while since I have looked at the repo...

PR #314, opened 2018-01-16
Add method to get item name

9a4gl:
Request handler might need this field. In my application javascript provide some extra info in name field and then server user that info.


src/ESPAsyncWebServer.h
```diff
@@ -267,6 +267,7 @@ class AsyncWebServerRequest {
     const String& header(size_t i) const;        // get request header value by number
     const String& headerName(size_t i) const;    // get request header name by number
     String urlDecode(const String& text) const;
+    const String& getItemName() const { return _itemName; }
```
me-no-dev: this is not something that is supposed to be accessed from the outside because there can be more than one item.

me-no-dev:
Left a note there :) sorry can not merge it. Use the other API to list parameters and their properties.

PR #334, opened 2018-03-07
Fix bug in const String& AsyncWebServerRequest::arg(const __FlashStringHelper * data) const

CarlosRDomin:
Fixes a crash when calling `AsyncWebServerRequest::arg` with a flash string param (eg: `request->arg(F("<arg name>"))` crashes). Solution is to call `strcpy_P` instead of `strcpy`, since the variable `p` is of type `PGM_P` and not `char *`.


me-no-dev:
Sorry for delaying this so long :)

PR #338, opened 2018-03-18
Handle multiple WebSocket frames within a TCP packet

adapt0:
This PR adds support for handling multiple WebSocket frames within a TCP packet.

Have been encountering an issue on Chrome 64.0.3282.186 where WebSocket message requests are missing during initial page load. Typically when there are a number of HTTP requests in flight.

Can reproduce by having the browser buffer WebSocket frames before sending:
```
ws = new WebSocket('ws://192.168.1.149/api/v1');
ws.onopen = function() {
  ws.send('#'.repeat(10000));
  for (var i = 0; i < 10; ++i) ws.send(i.toString());
}
```

Expect: bunch of fragmented frames followed by 10 short messages
Actual: bunch of fragmented frames

Frames are silently discarded when the received frame length is no longer a function of the received packet size. This PR adds a loop and adjusts the length checks to take into account that multiple frames may exist within a TCP packet.

### DIFF
--- a/src/AsyncWebSocket.cpp
+++ b/src/AsyncWebSocket.cpp
@@ -565,83 +565,93 @@ void AsyncWebSocketClient::_onDisconnect(){
   _server->_handleDisconnect(this);
 }
 
-void AsyncWebSocketClient::_onData(void *buf, size_t plen){
+void AsyncWebSocketClient::_onData(void *pbuf, size_t plen){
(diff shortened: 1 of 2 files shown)


me-no-dev:
Hey! Thanks for doing this! Really sorry that I have not had the time to take care of this repo recently

PR #366, opened 2018-05-19
Allow to override default template placeholder value with #define during compilation

me21:
There was a number of complaints from library users who needed to use the symbol of template placeholder in their data as-is. It seems that doubling the symbol (`%%` to get a single `%` in the output) doesn't work. While that should be fixed, the code of template engine is convoluted, so the proper fix might take some time.

This is a simple workaround: if the user #defines `TEMPLATE_PLACEHOLDER` during build, its value will be used. If not, default value of `%` is used.

### DIFF
--- a/src/WebResponseImpl.h
+++ b/src/WebResponseImpl.h
@@ -59,7 +59,10 @@ class AsyncAbstractResponse: public AsyncWebServerResponse {
     virtual size_t _fillBuffer(uint8_t *buf __attribute__((unused)), size_t maxLen __attribute__((unused))) { return 0; }
 };
 
+#ifndef TEMPLATE_PLACEHOLDER
 #define TEMPLATE_PLACEHOLDER '%'
+#endif
+
 #define TEMPLATE_PARAM_NAME_LENGTH 32
 class AsyncFileResponse: public AsyncAbstractResponse {
   using File = fs::File;


me21:
Er, the build error seems to have nothing to do with the changes.

JamesNewton:
I would really love to see this pulled.

JamesNewton:
I have a concern, confusion about this. Does the Arduino IDE re-compile a library if a define related to the library changes in the main .INO? I don't think it will. Triggering a re-compile of a library is a royal pain, as far as I can tell. Even just changing the library src doesn't update the code until you trackdown the arduinobuild folder and delete the .o file, right? So this isn't going to work. The only way to fix this is to make the template character a variable, which then needs a user interface to set it to other than the default. I hope I'm wrong here.

me21:
The #define should change not in the main .INO file, it should be passed as -D flag to the compiler during build. Or #defined directly in the library header file, but that's less suitable, I think. Both ways should trigger library recompiling.

zuqualla:
still a relevant thread! i solve it for now by putting my css in a separate file (without templates). 

Putting the #define TEMPLATE_PLACEHOLDER  in my main code doesn't work. Due to the scope of #define?

Changing it in the library itself is annoying when it gets updated.

me21:
You need to define that symbol as a build flag passed to the compiler. That's not easy to do in Arduino, but trivial in more sophisticated build systems (Platformio, etc.)

zuqualla:
thanks for your reply!
it's not a nice solution though, cause if we'd do this for every library we use it would become messy. 

Why not pass it as a argument? something like `server.templatePlaceholder = "$"`. 

I wish i had the skills to change it here on github, but still learning. So have to wait till one of you change it :)

ps: Platformio seems to have sold there soul to Microsoft, so for me that's a nogo. (i know about Atom, but it's looks they are moving away from it)

me21:
I'm afraid this repo is not actively maintained, at least currently :-(. So doing it yourself may be quicker.
Besides, this pull request is already merged. I think you should create new issue.

PR #376, opened 2018-07-02
Made example work in ESP32 and ESP8266

lbernstone:


me-no-dev:
I don't get it... why is travis failing?

me-no-dev:
does it compile on your side?
complains about SPI in some weird wifi lib

me-no-dev:
@lbernstone want to have another crack at this?

PR #380, opened 2018-07-10
Add handlers by URL templates

rjlexx:
This change allows add handlers by URL templates. 
For example:
server.on("/gpio*", handleGpio);
server.on("/irsend*", handleIrsend);
The "handleGpio" handler will handle all requests starting with "/gpio" and manage all GPIOs.
The "handleGpio" handler will handle all requests starting with "/irsend" and forward commands to IR manager.

### DIFF
--- a/src/WebHandlerImpl.h
+++ b/src/WebHandlerImpl.h
@@ -83,7 +83,13 @@ class AsyncCallbackWebHandler: public AsyncWebHandler {
       if(!(_method & request->method()))
         return false;
 
-      if(_uri.length() && (_uri != request->url() && !request->url().startsWith(_uri+"/")))
+      if (_uri.length() && _uri.endsWith("*")) {
+        String uriTemplate = String(_uri);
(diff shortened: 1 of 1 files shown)


src/WebHandlerImpl.h
```diff
@@ -83,7 +83,13 @@ class AsyncCallbackWebHandler: public AsyncWebHandler {
       if(!(_method & request->method()))
         return false;
 
-      if(_uri.length() && (_uri != request->url() && !request->url().startsWith(_uri+"/")))
+      if (_uri.length() && _uri.endsWith("*")) {
+        String uriTemplate = String(_uri);
+        uriTemplate.replace("*", "");
```
me-no-dev: Use `String::substring()` instead of replace

me-no-dev: it's the last symbol so it is easy :)

rjlexx: Ok. Thanks for the remark!

PR #391, opened 2018-07-31
Allow override rewrite match

WallyCZ:
This PR allows override rewrite matching so you can implement more powerfull rewrites. One small imperfect example (it rewrites f.e. "/radio/{frequence}" -> "/radio?f={frequence}"):

```cpp
class OneParamRewrite : public AsyncWebRewrite
{
  protected:
    String _urlPrefix;
    int _paramIndex;
    String _paramsBackup;

  public:
  OneParamRewrite(const char* from, const char* to)
    : AsyncWebRewrite(from, to) {

      _paramIndex = _from.indexOf('{');

      if( _paramIndex >=0 && _from.endsWith("}")) {
        _urlPrefix = _from.substring(0, _paramIndex);
        int index = _params.indexOf('{');
        if(index >= 0) {
          _params = _params.substring(0, index);
        }
      } else {
        _urlPrefix = _from;
      }
      _paramsBackup = _params;
  }

  bool match(AsyncWebServerRequest *request) override {
    if(request->url().startsWith(_urlPrefix)) {
      if(_paramIndex >= 0) {
        _params = _paramsBackup + request->url().substring(_paramIndex);
      } else {
        _params = _paramsBackup;
      }
    return true;

    } else {
      return false;
    }
  }
};
```

Use:

```cpp
  server.addRewrite( new OneParamRewrite("/radio/{frequence}", "/radio?f={frequence}") );
```

### DIFF
--- a/src/ESPAsyncWebServer.h
+++ b/src/ESPAsyncWebServer.h
@@ -303,11 +303,13 @@ class AsyncWebRewrite {
         _toUrl = _toUrl.substring(0, index);
       }
     }
+    virtual ~AsyncWebRewrite(){}
     AsyncWebRewrite& setFilter(ArRequestFilterFunction fn) { _filter = fn; return *this; }
     bool filter(AsyncWebServerRequest *request) const { return _filter == NULL || _filter(request); }
     const String& from(void) const { return _from; }
     const String& toUrl(void) const { return _toUrl; }
     const String& params(void) const { return _params; }
+    virtual bool match(AsyncWebServerRequest *request) { return from() == request->url() && filter(request); }
 };
 
 /*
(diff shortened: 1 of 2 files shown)


me-no-dev:
Wow :D almost a year later... I am adding this :) it's nice!

gusantor:
Hello & thank you !

I'm trying to use this, copied OneParamRewrite class on my sketch and added server.addRewrite( ...

But receive error message when compile

  'bool OneParamRewrite::match(AsyncWebServerRequest*)' marked override, but does not override

Please any guide on get this to work

RomeHein:
Hey! This would be nice to have some more informations/documentation on how to implement this.
This feature is important as it allows us to build restful api without having to play with build flags (regex).
I'm trying to add rewrite but this does not seem to work. 

Here is my implementation:
```
server.on("/gpio/scan", HTTP_GET, [this](AsyncWebServerRequest *request) { handleScan(request); });
server.addRewrite(new OneParamRewrite("/gpio/{pin}/scan", "/gpio/scan?p={pin}"));
```
`OneParamRewrite` class is the same than the exemple given above.
When doing this, my root always respond an `error 500`. When I remove the `addRewrite` line, the same root ("/gpio/scan") works fine. 
Any idea @WallyCZ ?

PR #407, opened 2018-09-02
Add method to check queue if is full

luc-github:
to control if can write or not and avoid the Error "too many messages"

### DIFF
--- a/src/AsyncWebSocket.cpp
+++ b/src/AsyncWebSocket.cpp
@@ -544,6 +544,12 @@ void AsyncWebSocketClient::_runQueue(){
   }
 }
 
+bool AsyncWebSocketClient::queueIsFull(){
+
+  if((_messageQueue.length() > WS_MAX_QUEUED_MESSAGES) || (_status != WS_CONNECTED) ) return true;
+  return false;
+}
+
 void AsyncWebSocketClient::_queueMessage(AsyncWebSocketMessage *dataMessage){
   if(dataMessage == NULL)
     return;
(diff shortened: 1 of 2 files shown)


src/AsyncWebSocket.cpp
```diff
@@ -876,6 +882,22 @@ void AsyncWebSocket::_handleDisconnect(AsyncWebSocketClient * client){
   });
 }
 
+bool AsyncWebSocket::availableForWriteAll(){
+  bool res = true;
+  for(const auto& c: _clients){
+    if(c->queueIsFull())res =false;
```
me21: It is possible here to return false immediately

src/AsyncWebSocket.cpp
```diff
@@ -876,6 +882,22 @@ void AsyncWebSocket::_handleDisconnect(AsyncWebSocketClient * client){
...
+  }
+  return res;
+}
+
+bool AsyncWebSocket::availableForWrite(uint32_t id){
+  bool res = true;
+  for(const auto& c: _clients){
+    if(c->queueIsFull() && (c->id() == id ))res =false;
```
me21: Same here, possible to return false immediately.

src/AsyncWebSocket.cpp
```diff
@@ -544,6 +544,12 @@ void AsyncWebSocketClient::_runQueue(){
   }
 }
 
+bool AsyncWebSocketClient::queueIsFull(){
+
+  if((_messageQueue.length() > WS_MAX_QUEUED_MESSAGES) || (_status != WS_CONNECTED) ) return true;
```
me21: Shouldn't it be `_messageQueue.length() >= WS_MAX_QUEUED_MESSAGES`?

src/AsyncWebSocket.cpp
luc-github: what you mean ?

me21: I mean you can write `return false;` instead of `res = false;`. That way you won't have to cycle through the remaining clients.

luc-github: Crap - what I did ? I should not stop drinking coffee these days...  anyway I have no excuse - this is not efficient code
Thank you for pointing it - I will redo PR properly

luc-github: redo https://github.com/me-no-dev/ESPAsyncWebServer/pull/410 and also changed condition to meet limit to prevent

PR #420, opened 2018-10-18
Update README.md ESP32 crash warning

hostmit:
There must be a warning to ESP32 users regarding crashes when handling multiple requests at some time.
Discussion is here:
[https://github.com/me-no-dev/ESPAsyncWebServer/issues/324]


me-no-dev:
fixed :)

Issue #390, opened 2018-07-30
Unable to include AsyncJson in different locations


dskw:
Since d7399a7664bce76e9939837fd6cf51bd272ef588 I am unable to have an include to AsyncJson in different files and run into the following linking error:
```
Archiving built core (caching) in: /tmp/arduino_cache_751141/core/core_esp8266_esp8266_nodemcuv2_CpuFrequency_80,FlashSize_4M1M,LwIPVariant_v2mss536,Debug_Disabled,DebugLevel_None____,FlashErase_none,UploadSpeed_115200_e35129e817913bbe969ad3c4750c592e.a
sketch/GameEditor.cpp.o:(.data.JSON_MIMETYPE+0x0): multiple definition of `JSON_MIMETYPE'
sketch/GameBoard.ino.cpp.o:(.data.JSON_MIMETYPE+0x0): first defined here
sketch/GameManager.cpp.o:(.data.JSON_MIMETYPE+0x0): multiple definition of `JSON_MIMETYPE'
sketch/GameBoard.ino.cpp.o:(.data.JSON_MIMETYPE+0x0): first defined here
```

The mentioned cpp heads look like this:
GameManager.cpp:
```
#include "GameManager.h"
#include "Utils.h"
#include <AsyncJson.h>
#include <ArduinoJson.h>
```

GameEditor.cpp:
```
#include "GameEditor.h"
#include "Utils.h"
#include <Hash.h>
#include <AsyncJson.h>
#include <ArduinoJson.h>
```

**Proposed solution:**
In AsyncJson.h change from this:
`const char* JSON_MIMETYPE = "application/json";`
to this:
`constexpr char* JSON_MIMETYPE = "application/json";`

tfelix:
I can confirm exactly the same error. I am no C++ expert but it seems the string inclusion in the header is done the wrong way (see: https://stackoverflow.com/questions/2879548/is-it-appropriate-to-set-a-value-to-a-const-char-in-the-header-file) and will break if the AsyncJson.h is included multiple times (which can easy happen if you for example use libraries like Basecamp for the ESP32).

Using your tip to works but gives me a warning: ISO C++ forbids converting a string constant to a char* 
So I vote for `const char* const JSON_MIMETYPE = "application/json";` which works fine.

0xC0DEBA5E:
[...]

PR #438, opened 2018-12-04
Fix compiler warning for AsyncJson’s JSON_MIMETYPE variable type.

buddydvd:
This commit fixes the following compiler warning:

    In file included from src/Program.cpp:21:0:
    lib/ESPAsyncWebServer/src/AsyncJson.h:39:33: warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]
    constexpr char* JSON_MIMETYPE = "application/json";

@tfelix had mentioned about this issue an earlier— see: https://github.com/me-no-dev/ESPAsyncWebServer/issues/390#issuecomment-409378991


buddydvd:
Here's more context:

When `JSON_MIMETYPE` was *first* introduced in https://github.com/me-no-dev/ESPAsyncWebServer/commit/d7399a7664bce76e9939837fd6cf51bd272ef588, it was defined as a _non-const_ global pointer to a `const char`:

    const char* JSON_MIMETYPE = "application/json";

Non-const global variables have external linkage. So, if you include `AsyncJson.h` in two different implementation files, you get the multiple definitions error. This is what #390 was about.

PR #396 (implementing the proposed solution described in #390) defined `JSON_MIMETYPE` as follow:

    constexpr char* JSON_MIMETYPE = "application/json";

This change fixes the redefinition error since `constexpr` specifier turned `JSON_MIMETYPE` a const global variable. _Const global variables in C++ have internal linkage by default_— so no multiple definitions error!

However, it may appear `JSON_MIMETYPE` still defines a pointer that points to a `const char` (when in fact, it doesn't). The change turned `JSON_MIMETYPE` from a _mutable-pointer to-a-const-char_ to a _const-pointer-to-a-mutable-char_. It's like as it were defined as follow:

    char* const JSON_MIMETYPE = "application/json";

Assigning the string literal `application/json` (`const char`) to a pointer that points to a mutable `char` is the reason for the compiler warning.

This PR restores the type that the `JSON_MIMETYPE` pointer points to (i.e., from `char` back to `const char`).

PR #439, opened 2018-12-04
Fix compiler warnings for AsyncJson’s _contentLength/_maxContentLength vars

buddydvd:
This commit fixes the following compiler warning:

    In file included from src/Program.cpp:21:0:
    lib/ESPAsyncWebServer/src/AsyncJson.h: In member function 'virtual void AsyncCallbackJsonWebHandler::handleBody(AsyncWebServerRequest*, uint8_t*, size_t, size_t, size_t)':
    lib/ESPAsyncWebServer/src/AsyncJson.h:153:64: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
    if (total > 0 && request->_tempObject == NULL && total < _maxContentLength) {

The type of `AsyncCallbackJsonWebHandler`'s member variables `_contentLength` and `_maxContentLength` should both be unsigned integer type `size_t`.


me-no-dev:
better late than never i guess :D

Issue #461, opened 2019-01-08
Corrupted websocket data + reboot using example code


tripplet:
Hi,

### Problem

The recevied websocket data by the ESP seems to be corrupted when it is send in multiple frames by the browser.
There seems to be same "e" at the end of frames which does not belong there and the data is corrupted in the 3. frame. After that the ESP reboots.

### Steps to reproduce

- I used the file server example from this repository [ESP_AsyncFSBrowser.ino](https://github.com/me-no-dev/ESPAsyncWebServer/blob/a7c4dfb04fd6a06001d9f87ba9935cc67cb256d5/examples/ESP_AsyncFSBrowser/ESP_AsyncFSBrowser.ino) with PlatformIO.
- No modification only the wifi name and password were changed.
- platformio.ini file: 
```ini
[env:nodemcuv2]
platform = espressif8266
board = nodemcuv2
framework = arduino
upload_speed = 921600
monitor_speed = 115200
lib_deps = https://github.com/me-no-dev/ESPAsyncWebServer.git
```
- Wait for the nodemcu to connect to the wifi and browse to the IP
  http://<i></i>192.168.0.113/heap

- Using the chrome dev tools executed the following JavaScript code:
```javascript
var ws = new WebSocket("ws://" + location.host + "/ws");
ws.onmessage = function(evt) { console.log(evt.data) };
```
- Wait for the `Hello Client 1 :)` response
- Execute the JavaScript code:
```javascript
[...]

Issue #457, opened 2019-01-05
ESP_AsyncFSBrowser.ino - bug notification


Pako2:
First of all, I would like to say that this library is my favorite - it's really great.
However - I have encountered some flaws in your example "ESP_AsyncFSBrowser". This works well **ONLY** if the "**the whole message is in a single frame and we got all of it's data**". Otherwise, it gives poor results:
1) The recipient can never receive the entire message. This is because the `onWsEvent` function is called multiple times, but the message buffer is reset every time (in line 28).
2) Even every single part of the message (received on one call `onWsEvent`) is not quite right. This is because "**info->len**" is incorrectly used instead of plain "**len**" in the loops on lines 61 or 66.

I have tried to solve these problems. The result is an attachment file. But I'm not a good programmer, so there are some bugs too.
[ESP_AsyncFSBrowser.zip](https://github.com/me-no-dev/ESPAsyncWebServer/files/2729212/ESP_AsyncFSBrowser.zip)

Pako2:
[This fix](https://github.com/me-no-dev/ESPAsyncWebServer/pull/462) is not entirely sufficient.
My comment - point 1 still applies.

PR #462, opened 2019-01-11
Fix example code

tripplet:
Based on Pako2s comment
See #457 and #461

### DIFF
--- a/examples/ESP_AsyncFSBrowser/ESP_AsyncFSBrowser.ino
+++ b/examples/ESP_AsyncFSBrowser/ESP_AsyncFSBrowser.ino
@@ -58,12 +58,12 @@ void onWsEvent(AsyncWebSocket * server, AsyncWebSocketClient * client, AwsEventT
       Serial.printf("ws[%s][%u] frame[%u] %s[%llu - %llu]: ", server->url(), client->id(), info->num, (info->message_opcode == WS_TEXT)?"text":"binary", info->index, info->index + len);
 
       if(info->opcode == WS_TEXT){
-        for(size_t i=0; i < info->len; i++) {
+        for(size_t i=0; i < len; i++) {
           msg += (char) data[i];
         }
       } else {
         char buff[3];
-        for(size_t i=0; i < info->len; i++) {
+        for(size_t i=0; i < len; i++) {
           sprintf(buff, "%02x ", (uint8_t) data[i]);
           msg += buff ;
         }


Pako2:
This fix is not entirely sufficient.
[My comment](https://github.com/me-no-dev/ESPAsyncWebServer/issues/457#issue-396135807) - point 1 still applies.

tripplet:
Your are right only fixes point 1.

me-no-dev:
@Pako2 you are supposed to get it in chunks. We can not store the whole frame :)

Pako2:
> @Pako2 you are supposed to get it in chunks. We can not store the whole frame :)

This is, of course, true, but I am still confident that the example demo contains a serious bug:
1) Line 28 - buffer "msg" is cleared here.
2) Line 62 or line 68 - you are "adding" another chunk into the buffer, **but that had been previously cleared and therefore any previous contents are lost!**

I tried to solve it in my modified example as you can see there. I used "client -> _ tempObject" instead of "msg" so I am not losing previous content when I receive another chunk.

me-no-dev:
it is not a flaw :) it's how Async is meant to be used. You get ONLY the data in the particular event. What you do with it, it's up to you. Store it in the _tempObject or elsewhere :) I can not add frame store to the example, because the first time that someone sends some large data to try things out and.... instant epic crash.

Pako2:
I understand you. The example actually works - the user can see the whole message (printed chunk by chunk).
However, I think that printing parts of a message (and then losing it) is usually quite useless. **Normally, the user needs the entire message for further processing.** And in the example, it's not clear how to do it (correctly - with reference to the client).
Perhaps it could at least be mentioned in some commentary.
But I understand your reasons. Too large a message can cause a problem and you need the example to work always and under all conditions ...

me-no-dev:
@Pako2 no need for bold text. You are the one that does not get it :) There is a big warning in the readme that async is not for the faint  hearted :) **This is the intended behaviour!** it is up to you to collect the whole message. The library has given you all possible information that you may need to do that. Wether you like it or not is irrelevant. This is how this library has always been meant to work. The example is there to show you how the data comes and how it is organised. From then on, do as you wish with it. The library and code as-is can handle ANY size packet you throw at it. Your suggestion will make it handle only small packets and frames.

me-no-dev:
Think of it as kinda deterrent for noobs. Though there are so many examples and uses out there now that all kinds of people run async. I had to decide if I want to make the library easy to use, or to be able to push the limits with it. I decided the later. There is the regular web server in Arduino for the rest. I myself struggle sometimes to implement some things in Async also ... :)  it's not an easy topic. I also rarely have time to spend on personal projects (it has been probably a year since I last spent a day on async...) and I do not want the issue tracker to blow (though that heap issue did it well). I realise that I might sound like a not-so-nice person, but I am :) just trying to make things easy for all :D (that can handle async).

PR #474, opened 2019-02-17
Adjust canHandle() for AsyncEventSource

kaegi:
Neither `curl` nor a web browser can access an event stream because they do not send `expected request connect types` (so the event source handler is skipped). I assume the way `curl` and web browsers work is standard, so we can simply remove the condition `request->isExpectedRequestedConnType(RCT_EVENT)`.


me-no-dev:
OK :) let's see how it will do. Was there to cover specs and what I have observed.

PR #478, opened 2019-02-20
Modification to the embedded SPIFFS editor, to allow loading the main…

wkraft-fablabka:
… JS-editor from SPIFFS instead from PROGMEM.

This was done renaming the SPIFFSEditor class to SPIFFSEditorBase, which creates the response from SPIFFS.
A newly created derived class SPIFFSEditor behaves like before, so the change is backward compatible.

Subclassing was used -instead of a single if-clause- to make sure, that the linker does not tie the 4k PROGMEM blob into the executable, when it is not used.

Loading the editor JS object from SPIFFS allows much easier updates and changes of the Javascript parts, without recompiling the C-Code after generating hex arrays of minimized HTML.


wkraft-fablabka:
travis is still crashing :-(

me-no-dev:
This is one way of doing things :) or you can just add a class variable and methog that allows you to change the default index. That way you could do it on the fly if you upload one later :)

PR #483, opened 2019-02-26
change class members from private to protected, so subclasses are abl…

wkraft-fablabka:
…e to access client information


me-no-dev:
what is the use case?

PR #491, opened 2019-03-17
arduino json 6 async web server migration

baggior:
fix for ArduinoJson v. 6

PR #501, opened 2019-04-11
Detect if AsyncWebSocketClient can send data to server and add AsyncWebServer::end() 

dyarkovoy:
If send queue overflows, the messages sent by AsyncWebSocketClient are discarded and the caller is not able to determine whether the send was (or will be) successful. 
Suggested workaround is to add AsyncWebSocketClient::canSend() method that can be called to determine if the next add to the sending queue will succeed.
#504 should be easily fixable with this, too.
This PR also adds AsyncWebServer::end() in case we ever want to stop Web server. Currently it doesn't seem to be feasible


baggior:
my pull request #491 contains fixes for travis

me-no-dev:
Looks nice :) Sorry for the delayed reply. Please rebase so I can merge it :)

dyarkovoy:
> Looks nice :) Sorry for the delayed reply. Please rebase so I can merge it :)

Thanks, you should be able to merge now

me-no-dev:
BTW I see that you compare less or equal, which means that it will return true when the queue is actually full?

me-no-dev:
is that the intended behaviour?

dyarkovoy:
> is that the intended behaviour?

it is based on this piece in AsyncWebSocketClient::_queueMessage():

>   if(_messageQueue.length() > WS_MAX_QUEUED_MESSAGES){
>       ets_printf("ERROR: Too many messages queued\n");
>       delete dataMessage;
>   } else {
>       _messageQueue.add(dataMessage);
>   }
> 

I figured it will still add when the length()==WS_MAX_QUEUED_MESSAGES

me-no-dev:
maybe wrong on both places? my logic dictates that if you are more or equal to the queued messages, then you should fail. Isn't that correct?

dyarkovoy:
agreed 100%. Just didn't want to make more changes than necessary. Do you want me to make the changes in both places?

me-no-dev:
yes please :)

dyarkovoy:
done!

me-no-dev:
and done :) thanks!

PR #509, opened 2019-05-08
Add methods to determine Average Queue length

philbowles:
When sending events rapidly, the internal queue grows if ratein > rateout. Unless the sender has a method of knowing when this is happening, he/she can easily overload the queue, causing rapid heap depletion and (especially in the low-memory ESP8266) an uncontrolled crash. An arbitrary fixed-length queue is not a workable solution, especially without a feedback mechanism to tell when messages have been discarded.

Proposed changes allow user to query the size of the Q and then "back off" / pause / wait before sending more messages and in doing so, prevent an uncontrolled crash.

Since the client holds the Q it is modified to simply return its length. A connection-wide routine then averages the Q lengths of all the clients to return a single value showing the "busyness" / loading / imminent crash status of the library.

Code has been in use for several weeks with zero side effects other than preventing uncontrolled crashes of the library.


me-no-dev:
Hey :) How about you make the names self explanatory? like `packetsWaiting` or something. not sure, but MQL... :)

me-no-dev:
also please rebase so the build passes.

philbowles:
Fair comment re the names that was just me testing it...also I'm a bit new @ github so I don't even know what "also please rebase so the build passes." means. I will address both issues

PR #531, opened 2019-06-22
Fix sha1 not defined compilation error

gmag11:
With ESP8266 Arduino core 2.5.2 I was getting a compilation error related to sha1() function. Hash.h was not included.
After adding it to AsyncWebSocket.h code compiles fine.
I'm using Visual Studio 2019 with VisualMicro 1906.16.1


me-no-dev:
Sure :)

PR #398, opened 2018-08-19
 json response is more human readable with PrettyAsyncJsonResponse class

gerald-guiony:
json response example 
 

{
  "deviceName": "ESP588624",
  "chipId": 588624,
  "heap": 24032,
  "lastResetReason": "Power on",
  "coreVersion": "2_3_0",
  "sdkVersion": "1.5.3(aec24ac9)",
  "cpuFreq": 80,
  "sketchSize": 430528,
  "freeSketchSpace": 614400,
  "freeSpiffsSpace": true,
  "accessPoint": false,
  "IP": "192.168.1.15",
  "macAddress": "60:01:94:08:FB:50",
  "subnetMask": "255.255.255.0",
  "gatewayIP": "192.168.1.254",
  "dnsIP": "192.168.1.254",
  "hostname": "ESP588624",
  "SSID": "freebox_ASUTRY",
  "BSSID": "E4:9E:12:B7:5D:34",
  "RSSI": -75
}

PR #533, opened 2019-06-24
Update AsyncJson.h to fix ArduinoJson v6

baggior:
for the newest ArduinoJson v6 compatibility


baggior:
needed to fix the compilation bug introduced in #398 when using Arduino Json v6

this fix simply disables the pretty print class in that case...

thbl:
@baggior do you have an example on how to use this ?
As example in readme dosen't work with v6
Thanks

PR #561, opened 2019-07-23
Add support for content type with charset

Bmooij:
Add support for Content-Type with charset

Content-Type: text/html; charset=utf-8

_It is ignoring the charset_

### DIFF
--- a/src/WebRequest.cpp
+++ b/src/WebRequest.cpp
@@ -316,13 +316,11 @@ bool AsyncWebServerRequest::_parseReqHeader(){
     if(name.equalsIgnoreCase("Host")){
       _host = value;
     } else if(name.equalsIgnoreCase("Content-Type")){
+	  _contentType = value.substring(0, value.indexOf(';'));
       if (value.startsWith("multipart/")){
         _boundary = value.substring(value.indexOf('=')+1);
         _boundary.replace("\"","");
-        _contentType = value.substring(0, value.indexOf(';'));
         _isMultipart = true;
-      } else {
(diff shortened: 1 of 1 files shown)


me-no-dev:
charset should be a separate variable :) it's not expected by the rest of the code to get such content type and some parsing may/will fail

Bmooij:
I walked into this problem when writing a flutter app. 
[Remove charset from "Content-Type" header](https://github.com/dart-lang/http/issues/184)

According to RFCs and it seems the "parameter" part of the Content-Type value is either mandatory or optional depending on the particular MIME type/subtype:
[https://tools.ietf.org/html/rfc2045#section-5](https://tools.ietf.org/html/rfc2045#section-5)

src/WebRequest.cpp
```diff
@@ -310,13 +310,11 @@ bool AsyncWebServerRequest::_parseReqHeader(){
     if(name.equalsIgnoreCase("Host")){
       _host = value;
     } else if(name.equalsIgnoreCase("Content-Type")){
+	  _contentType = value.substring(0, value.indexOf(';'));
```
atanisoft: what happens if there is no ';' in the value? A check should likely be made to assign the substring only if there is a ';' mark otherwise use the value as-is.

Bmooij: If `value.indexOf` returns -1 (not found)[https://www.arduino.cc/reference/en/language/variables/data-types/string/functions/indexof/]

`value.substring` returns the complete string

PR #580, opened 2019-08-24
Add function so that the total number of web-socket clients can be li…

matt123p:
This is to cope with a problem when a browser does not close the web-socket connection correctly.  

I have observed this on Chromium based browsers.  The resource leak will eventually lead to the server crashing with random heap errors (after around 13 or so client connections).  

Normally only one connection per client is required, so limiting the number of connections would not normally cause any problems.

The new function checks to see if the maximum number concurrent connections is exceeded and then closes the oldest connection.  If this connection was required then the client will normally just reconnect.

It should be called from with in the main loop periodically.  In my application I call it around once per second and this seems to work fine.  A really obnoxious client could crash your server by opening connections faster but it seems unlikely that would be a problem in most real-world applications.


matt123p:
Wrong branch, I will re-submit

PR #585, opened 2019-08-29
Add thread locking to improve stability.

matt123p:
If you use web-sockets you get frequent heap corruption errors.  By a process of trail and error I have tracked this down to the websocket buffer.

As far as I can tell the websocket class is access by 2 different threads, the application thread and the tcpip thread.  Normally this is fine because thread synchronisation is handled by message queues.  

However, the websocket buffer linked list is used by both threads and a race condition exists which will cause heap corruption if a buffer is being created by the application when at the same time one is being consumed by the tcpip thread.

This patch adds thread safety to the websocket buffer linked list.  It shouldn't add too much overhead to the websocket implementation as the locks are only in place for a relatively short time and they are only used on construction and destruction of a websocket buffer.

### DIFF
--- a/src/AsyncWebSocket.h
+++ b/src/AsyncWebSocket.h
@@ -31,6 +31,8 @@
 #endif
 #include <ESPAsyncWebServer.h>
 
+#include "AsyncWebSynchronization.h"
+
 #ifdef ESP8266
 #include <Hash.h>
 #endif
@@ -236,6 +238,7 @@ class AsyncWebSocket: public AsyncWebHandler {
     uint32_t _cNextId;
     AwsEventHandler _eventHandler;
     bool _enabled;
+	AsyncWebLock _lock;
   public:
     AsyncWebSocket(const String& url);
     ~AsyncWebSocket();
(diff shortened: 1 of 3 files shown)


me-no-dev:
yup :) I am aware of all of this. Just do not have the time to code this properly for both esp32 and esp8266. You can notice that Travis is failing the build for esp8266, because it does not run FreeRTOS and it does not have semaphores ;)

me-no-dev:
Try to make this compatible with both by using defines and such ;) On esp8266 it can be a volatile bool or something.

matt123p:
Sure, no problem, I will re-submit when I have fixed the build for the esp8266.

matt123p:
Hi, I couldn't find a suitable atomic function to read and replace a boolean - perhaps you know one?

For now, I have removed the code for the 8266 but i am happy to revisit if somebody cam point me towards a suitable set of atomic functions.

me-no-dev:
I would say to use the defines only in the lock class where you call semaphores. You can leave a note and use a bool there for 8266 so I can have a look later. If you do it like that, then I can reuse the class easily elsewhere, without having to add 3 lines every time and clutter the code.

matt123p:
OK, I did as requested, and make two versions of the lock class, one for the ESP32 and one for the 8266.

src/AsyncWebSocket.h
```diff
@@ -236,6 +238,8 @@ class AsyncWebSocket: public AsyncWebHandler {
     uint32_t _cNextId;
     AwsEventHandler _eventHandler;
     bool _enabled;
+	  AsyncWebLock _lock;
```
me-no-dev: looks like extra tab made it here :)

me-no-dev:
looks nice :) just one style request

matt123p:
Ok. Done.

PR #591, opened 2019-09-09
Add function so that the total number of web-socket clients can be limited

matt123p:
This is to cope with a problem when a browser does not close the web-socket connection correctly.

I have observed this on Chromium based browsers. The resource leak will eventually lead to the server crashing with random heap errors (after around 13 or so client connections).

Normally only one connection per client is required, so limiting the number of connections would not normally cause any problems.

The new function checks to see if the maximum number concurrent connections is exceeded and then closes the oldest connection. If this connection was required then the client will normally just reconnect.

It should be called from with in the main loop periodically. In my application I call it around once per second and this seems to work fine. A really obnoxious client could crash your server by opening connections faster but it seems unlikely that would be a problem in most real-world applications.


me-no-dev:
Do you mind adding this to the examples as well? At least to the ones using WS. Also will be good to add a note in the README :)

matt123p:
I have updated the README and the web socket example source.  The build is now failing because the example is building against the unpatched library not the new one :(

matt123p:
OK.  Stupid error.  I have fixed the build.

src/AsyncWebSocket.h
```diff
@@ -255,6 +255,7 @@ class AsyncWebSocket: public AsyncWebHandler {
 
     void close(uint32_t id, uint16_t code=0, const char * message=NULL);
     void closeAll(uint16_t code=0, const char * message=NULL);
+    void cleanupClients(uint16_t maxClients = 4);
```
me-no-dev: this number should really equal the max open TCP connections of the device. for ESP8266 that is 4 or 5 and for ESP32 that is 16. Having it 4 by default will make it limited on ESP32

matt123p:
Remember this is is the number of web socket connections, not total TCP connections.  Typically you will only want one connection per client and you will want to keep some TCP slots available for the other HTTP connections.

Setting it to 4 means that the server will support a maximum of 4 simultaneous clients.  I found when the number of connections reaches around 12, then the ESP32 crashes - probably due to the fact there are still some connections open for the other HTTP fetches for the page.  If you let it go to the maximum of 16 connections then no other client will be able to connect to the server.

me-no-dev:
4 seems low though. let's agree on higher value for esp32

matt123p:
How about 8?  This is only the default the user can override it by passing a parameter.

me-no-dev:
8 sounds better :)

PR #605, opened 2019-09-26
added dependency for PlatformIO ldf_mode off

kleini:
When deactivating the PlatformIO dependency finder all dependencies of used libraries need to be defined properly.


library.json
```diff
@@ -24,6 +24,9 @@
     {
       "name": "AsyncTCP",
       "platforms": "espressif32"
+    },
+    {
+      "name": "Hash"
```
atanisoft: this needs to have ```"platforms" : "espressif"``` as this dependency is only used on the esp8266.

kleini: Yes, you're right. I saw in the code the #ifndef. Fixed.

PR #611, opened 2019-10-02
Fix canHandle

Bmooij:
I encountered the problem that the wrong function is called.

My code:
```
webserver->on("/system/reset", HTTP_PUT, std::bind(&MyClass::myFunc1, this, std::placeholders::_1));
webserver->on("/system/reset/factory", HTTP_PUT, std::bind(&MyClass::myFunc2, this, std::placeholders::_1));
```

When I call: `curl -X PUT "http://192.168.4.1/system/reset/factory`

`MyClass::myFunc1` is executed.

This was because of [startsWith](https://github.com/me-no-dev/ESPAsyncWebServer/blob/01019542c2e1fd484332c831f8f11e8369fd4649/src/WebHandlerImpl.h#L109)
```"/system/reset/factory".startsWith("/system/reset/") == true```

### DIFF
--- a/src/WebHandlerImpl.h
+++ b/src/WebHandlerImpl.h
@@ -106,7 +106,7 @@ class AsyncCallbackWebHandler: public AsyncWebHandler {
         if (!request->url().startsWith(uriTemplate))
           return false;
       }
-      else if(_uri.length() && (_uri != request->url() && !request->url().startsWith(_uri+"/")))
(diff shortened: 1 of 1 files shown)


me-no-dev:
this is intentional. in the docs I mention that handlers are evaluated in the order they are added. just swap the two lines and you will get what you want :) `/system/reset/factory` will be evaluated first and used. More specific handlers should be first

PR #615, opened 2019-10-03
Fixes #614 with incorrect regex in the example sketch

frippe75:
Incorrect escape of backslash in regex_patterns.ino sketch. Verified on one of my ESP32's

Issue #613, opened 2019-10-02
binary project size at 1.2.3 increases dramatically


reaper7:
today I updated ESPAsyncWebServer library to 1.2.3
and binary file for the espurna project increased its volume by about 100,000 bytes!

that's very strange because yesterday I was also updating the git version and everything was ok.

...so the same espurna project with ESPAsyncWebServer in the version from before today's changes:
```
IROM   : 458060          - code in flash         (default or ICACHE_FLASH_ATTR) 
IRAM   : 28308   / 32768 - code in IRAM          (ICACHE_RAM_ATTR, ISRs...) 
DATA   : 1440  )         - initialized variables (global, static) in RAM/HEAP 
RODATA : 5540  ) / 81920 - constants             (global, static) in RAM/HEAP 
BSS    : 27416 )         - zeroed variables      (global, static) in RAM/HEAP 

Użycie biblioteki ArduinoJson_5x w wersji 5.13.5 z folderu: C:\Programy\arduino_projekty\libraries\ArduinoJson_5x 
Użycie biblioteki ESP8266WiFi w wersji 1.0 z folderu: C:\Programy\arduino\hardware\esp8266com\esp8266\libraries\ESP8266WiFi 
Użycie biblioteki eeprom_rotate w wersji 0.9.2 z folderu: C:\Programy\arduino_projekty\libraries\eeprom_rotate 
Użycie biblioteki EEPROM w wersji 1.0 z folderu: C:\Programy\arduino\hardware\esp8266com\esp8266\libraries\EEPROM 
Użycie biblioteki ESPAsyncTCP w wersji 1.2.2 z folderu: C:\Programy\arduino_projekty\libraries\ESPAsyncTCP 
Użycie biblioteki async-mqtt-client w wersji 0.8.2 z folderu: C:\Programy\arduino_projekty\libraries\async-mqtt-client 
Użycie biblioteki ArduinoOTA w wersji 1.0 z folderu: C:\Programy\arduino\hardware\esp8266com\esp8266\libraries\ArduinoOTA 
Użycie biblioteki embedis w wersji 1.2.0 z folderu: C:\Programy\arduino_projekty\libraries\embedis 
Użycie biblioteki ESPAsyncWebServer w wersji 1.2.2 z folderu: C:\Programy\arduino_projekty\libraries\ESPAsyncWebServer 
[...]

PR #617, opened 2019-10-06
Make regex a define exclusive

Bmooij:
Make regex a define exclusive

Based on feedback: https://github.com/me-no-dev/ESPAsyncWebServer/pull/615#issuecomment-538722184
And: https://github.com/me-no-dev/ESPAsyncWebServer/issues/613

It is not finished/tested yet, but feedback is always welcome.


me-no-dev:
maybe have the public functions do nothing or return error if not enabled. and why is the define name change necessary?

me-no-dev:
checks are failing in the regex example

Bmooij:
> maybe have the public functions do nothing or return error if not enabled. 
I was struggling with that. 
On one side for code readability, I should say "return error if not enabled". (Run time error)
On the other side, if not enabled, you cannot use the corresponding functions. (Compile time error)

> why is the define name change necessary?
I was not sure of the name of the define.
So there is one place to change the name of the define.

Bmooij:
I reverted the performance improvement. 
Mentioned [here](https://github.com/me-no-dev/ESPAsyncWebServer/issues/613#issuecomment-537918572)

It uses (a lot) more heap, for to little performance  improvement (in my opinion)

.github/scripts/install-arduino-ide.sh
```diff
@@ -107,6 +108,7 @@ function build_sketch(){ # build_sketch <fqbn> <path-to-ino> [extra-options]
 		-libraries "$ARDUINO_USR_PATH/libraries" \
 		-build-cache "$ARDUINO_CACHE_DIR" \
 		-build-path "$ARDUINO_BUILD_DIR" \
+		-prefs=compiler.cpp.extra_flags="$build_flags" \
```
me-no-dev: this should maybe include the c flags as well, not just cpp

.github/scripts/install-arduino-ide.sh
```diff
@@ -210,7 +212,13 @@ function build_sketches() # build_sketches <fqbn> <examples-path> <chunk> <total
         || [ "$sketchnum" -gt "$end_index" ]; then
         	continue
         fi
-        build_sketch "$fqbn" "$sketch" "$xtra_opts"
+		local sketchBuildFlags=""
+		if [ -f "$sketchdir/.test.build_flags" ]; then
+			while read line; do
+				sketchBuildFlags="$sketchBuildFlags $line"
+			done < "$sketchdir/.test.build_flags"
+		fi
```
me-no-dev: please care for tabs/spaces. in other words... do not use tabs :)

README.md
```diff
@@ -1484,3 +1485,30 @@ webServer.onNotFound([](AsyncWebServerRequest *request) {
...
+```cpp
+  server.on("^\\/sensor\\/([0-9]+)$", HTTP_GET, [] (AsyncWebServerRequest *request) {
+      String sensorId = request->pathArg(0);
+  });
+```
+*NOTE*: All regex patterns starts with `^` and ends with `$`
+
+To enable the `Path variable` support, you have to define the buildflag `-DASYNCWEBSERVER_REGEX`.
+
+For Arduino IDE modify `platform.local.txt`:
```
me-no-dev: where is that platform.local.txt?

src/ESPAsyncWebServer.h
```diff
@@ -271,13 +273,20 @@ class AsyncWebServerRequest {
...
     const String& headerName(size_t i) const;    // get request header name by number
     String urlDecode(const String& text) const;
+
+#ifdef _ASYNCWEBSERVER_REGEX
+    private:
+      LinkedList<String *> _pathParams;
+      void _addPathParam(const char *param);
+
+    public:
+      const String& pathArg(size_t i) const;
```
me-no-dev: this functions should exists either way. just return whatever is the equivalent of error when regex is disabled.

me-no-dev: in other words... examples that use it, should compile regardless if regex is enabled or not.

Bmooij: Personally I'm not a fan of run time errors. Now you get a error on compile time when ASYNCWEBSERVER_REGEX was not defined.

But then again that is my opinion.

Bmooij: I've changed it. If ASYNCWEBSERVER_REGEX is not defined it handles the function if the pathArg was not found.

I added a compiler warning, if the function is used but ASYNCWEBSERVER_REGEX was not defined

README.md
Bmooij: I found out that platform.local.txt is not needed.

For ardruino IDE:
```cpp
#define ASYNC_WEBSERVER_REGEX
#include <ESPAsyncWebServer.h>
```

.github/scripts/install-arduino-ide.sh
Bmooij: Replaced all the tabs with spaces

.github/scripts/install-arduino-ide.sh
Bmooij: Currently I'm only using a CPPDEFINE

Bmooij:
> why change all tabs/spaces. change ONLY your edits. now I have no clue what is going on. Revert that and apply only on your changed lines.

Sorry for the inconvenience, I reverted it and applied it only on my added lines

README.md
Bmooij: I will switch back to the platform.local.txt method. 
I do not prefer that method, but it is the only method that works with the Arduino IDE

me-no-dev: yes, unfortunately we can not add menus from the libraries (or something like that). Arduino is quite more basic that most of the editors that I use. Users know to poke around. That is how ESP8266 has SSL support in ESPAsyncTCP

me-no-dev:
alright :) it looks really nice now. Thanks for adding docs ;) Will merge when the build finishes :)

PR #620, opened 2019-10-11
Remove warnings for unused params

devyte:
I'm getting compile warnings in my env for unused params. This removes those warnings.
I do wonder why these warnings aren't apparent in CI.


devyte:
C-style casts apparently aren't appreciated. Please advise on how to proceed.

me-no-dev:
i'm kinda unhappy with Codacy's decisions here... :P On top of that I can no longer find that requirement in CPP ;)

PR #621, opened 2019-10-15
AsyncEvents/ServerSideEvents: prevent internal DOS by prevent overflowing messageQueue

iafilius:
Hi,
First of all I think the AsyncEvents is not SMP safe, and crashes a lot when stressed a bit.

This pull request adresses only one issue, basically copied solution from the websockets code.
In case a SSE client is connected, and messages are generated rapidly, the ESP32 WiFi/tcpip stack will lockup when the wifi connection(distance) is getting worse. _messageQueue is completely filled up, basically denial of service to the TCP stack.
WDT goes off when not disabled.

I copied the include part from web sockets include, rename the WS -defines to SSE-defines, and changed queing code, now queue is filled up until a certain level, most recent messages are dropped above, instead wildly stressing tcp stack.

I see the web Sockets have a controll queue, which i think is required for serverside events (AsyncEvents) too , but do not know yet how to start on that. In meanwhile this patch solved the most annoying lockup of wifi/tcpstack (in my case)

### DIFF
--- a/src/AsyncEventSource.cpp
+++ b/src/AsyncEventSource.cpp
@@ -184,10 +184,14 @@ void AsyncEventSourceClient::_queueMessage(AsyncEventSourceMessage *dataMessage)
     delete dataMessage;
     return;
   }
-
-  _messageQueue.add(dataMessage);
-
-  _runQueue();
+  if(_messageQueue.length() >= SSE_MAX_QUEUED_MESSAGES){
+      ets_printf("ERROR: Too many messages queued\n");
+      delete dataMessage;
+  } else {
+      _messageQueue.add(dataMessage);
+  }
+  if(_client->canSend())
+    _runQueue();
 }
 
 void AsyncEventSourceClient::_onAck(size_t len, uint32_t time){
--- a/src/AsyncEventSource.h
+++ b/src/AsyncEventSource.h
@@ -23,10 +23,27 @@
 #include <Arduino.h>
 #ifdef ESP32
 #include <AsyncTCP.h>
+#define SSE_MAX_QUEUED_MESSAGES 32
 #else
 #include <ESPAsyncTCP.h>
+#define SSE_MAX_QUEUED_MESSAGES 8
 #endif
 #include <ESPAsyncWebServer.h>
+
+#include "AsyncWebSynchronization.h"
+
+#ifdef ESP8266
+#include <Hash.h>
+#ifdef CRYPTO_HASH_h // include Hash.h from espressif framework if the first include was from the crypto library
+#include <../src/Hash.h>
+#endif
+#endif
+
+#ifdef ESP32
+#define DEFAULT_MAX_WS_CLIENTS 8
+#else
+#define DEFAULT_MAX_WS_CLIENTS 4
+#endif
 
 class AsyncEventSource;
 class AsyncEventSourceResponse;
--- a/src/AsyncEventSource.h
+++ b/src/AsyncEventSource.h
@@ -40,9 +40,9 @@
 #endif
 
 #ifdef ESP32
-#define DEFAULT_MAX_WS_CLIENTS 8
+#define DEFAULT_MAX_SSE_CLIENTS 8
 #else
-#define DEFAULT_MAX_WS_CLIENTS 4
+#define DEFAULT_MAX_SSE_CLIENTS 4
 #endif
 
 class AsyncEventSource;


me-no-dev:
yup, there is more to do and I am already working on reworking the web sockets. The lib was not intended for RTOS/SMP and I just never got around to fix web sockets and event source, since those are the public interfaces that get called from other threads. The regular server part runs in one thread and is fine.

iafilius:
Thanks. 
With a rework on the way for websockets, and ServerSideEvents shares a lot, i expect not too wise to try to catch up worth current websockets myself.

A tip for those experiencing instability (in Events), 
defining:
#define CONFIG_FREERTOS_UNICORE
#define CONFIG_ASYNC_TCP_RUNNING_CORE 1
before other includes greatly improved stability (for me), by locking webserver and asynctcp both on same CPU.

Regards

philbowles:
There is already a fix for this! It is an absolute nightmare fixing the queue length at an arbitrary figure of 8. I had to stop using websockets after the"fix" rendered them completely unusable for anything serious.

As a result I changed to using SSE because they were not crippled by an absurdly low fixed value and I spent  a lot of time developing a fix to allow the user to query the queue length (see AsyncEventSource.cpp line 297 function size_t AsyncEventSource::avgPacketsWaiting() ) and make up his/her own mind what a sensible maximum value for his or her app.

The existing fix is flexible and allows the user to decide, this one is

1. too rigid
2. too low
3. removes user flexibility
4. duplicates (but almost breaks) an existing fix

This fix takes the is a huge retrograde step and seriously needs to be unwound!

@me-no-dev PLEASE PLEASE PLEASE revert this change!

me-no-dev:
@iafilius want to comment?

iafilius:
Hi,

My 5 cents.

As explained the workaround copied from web-sockets does it’s job for me perfectly to prevent a crash situation in high-rate/bad connectivity situation.
I do not really like/desire  the hard coded 'fix'/mask of the (real) problem, so can i guess perfectly use the avgPacketsWaiting(), or maybe a new totalPacketsWaiting() or thisClientPacketsWaiting() even better. Stability out of the box has my preference, hard coded or not.
I did not notice/pay attention to the function before,  a small description /sample would help, perhaps added with new totalPacketsWaiting and thisClientPacketsWaiting() as well

It looks to me websockets is active developed on, and seems leading in progress in relation to asyncEvents. most important some active developments on queue management.
As WS and SSE share a lot of functionality i vote to try to make or keep the implementations as much as possible the same, or similar.


Maybe:
1) Keep SSE_MAX_QUEUED_MESSAGES, but make it possible to overrule (or disable with set to -1 ) in own code (currently not but easy to do)
   So even simple examples always work and do not crash in high rate/slow connection case.
   But we should preferable do the same in web sockets to keep in sync.
2) I'd notice some active queue management code already in WebSockets, cleaning dead sessions as well. probably good to implement in SSE as well.
   to my opinion/experience the SSE queue isn't SMP proof and crashes quite easy, I would welcome any improvement on that, and current WebSockets looked like a move to the right direction.
   Me-No-Dev wrote about working on a new implementation for web sockets, and i hope for SSE as well.
3) Document avgPacketsWaiting() and implement a few sister functions as well like totalPacketsWaiting and thisClientPacketsWaiting(). (Similar for both SSE and WebSockets)
4) stop throwing away most recent message when "system limit" has been reached, but throw oldest into bin, it might be a matter of taste  or typical use-case. (both in SSE and websockets)



sample regarding #1
#ifndef SSE_MAX_QUEUED_MESSAGES
#define SSE_MAX_QUEUED_MESSAGES 32
#endif


change: 
  if (_messageQueue.length() >= SSE_MAX_QUEUED_MESSAGES) {
    ets_printf("ERROR: Too many messages queued\n");
    delete dataMessage;
  } else {
    _messageQueue.add(dataMessage);
  }

to:
  if (SSE_MAX_QUEUED_MESSAGES > 0 && _messageQueue.length() >= SSE_MAX_QUEUED_MESSAGES) {
    ets_printf("ERROR: Too many messages queued\n");
    delete dataMessage;
  } else {
    _messageQueue.add(dataMessage);
  }



custom code may define:
# 'crazy' queue limit:
#define SSE_MAX_QUEUED_MESSAGES 128
or to disable:
#define SSE_MAX_QUEUED_MESSAGES -1

and then for both websockets and SSE.

As it might become clear i would really see both SSE and Websockets to share code, queue management, fixes/workaorund etc.
Not letting both completely grow from each other and become separate beasts, and stable out of the box.


what do you think.

Regards,

iafilius:
Hi,
@me-no-dev @philbowles  

created untested demo code which allows to disable hardcoded queuelength, or set larger/smaller limit in user code based on concerns by Phil.
it is available at:
https://github.com/iafilius/ESPAsyncWebServer/tree/option_to_remove_SSE_or_WS_queuelen

It is untested (not ready for pull request), does the same logic adjustment in websockets as well in SSE.
With websockets the WS_MAX_QUEUED_MESSAGES  is used as well in queueIsFull() which is internally used at a few places, and I did not change that function/logic. i guess that would need attention as well.

Would path this take the concerns away??
shall i make a pull request for it? (but what to do with queueIsFull() )

PS: my vscode i currently used as editor likes to reformat the whole code, fine for my own code, but not really when contributing to a project (yes i know very small)

thanks

PR #660, opened 2019-12-11
Premature client close() will leak _itemBuffer

mcspr:
Easy way to reproduce is to have:
```
 server->on("/data", HTTP_POST, _onDataRequest, _onDataRequestFile);
```
And call `request->client()->close()` inside of the `onDataRequestFile()` callback after some time, while uploading ~1MB of data

I would assume this would also happen when remote closes the connection earlier than we could've parsed file contents

### DIFF
--- a/src/WebRequest.cpp
+++ b/src/WebRequest.cpp
@@ -96,6 +96,10 @@ AsyncWebServerRequest::~AsyncWebServerRequest(){
   if(_tempFile){
     _tempFile.close();
   }
+	
+  if(_itemBuffer){
+    free(_itemBuffer);
+  }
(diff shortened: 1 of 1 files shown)


zhivko:
Great @mcspr  It seems you found it!? huray!!! @me-no-dev ?

mcspr:
I also wonder if manual destructor stuff could be avoided altogether by replacing pointer members with unique_ptr. This would also allow us to use C++ objects.
LinkedList destructor is manual too, _headers, _params, _interestingHeaders all need .free(), but I am not sure if that is intentional for some reason.

PR #688, opened 2020-01-28
Handshake handlers

ayushsharma82:


ayushsharma82:
Added some handshake handlers for AsyncWebsocket and AsyncEventSource

ayushsharma82:
Tested to be working with ESP8266.

PR #697, opened 2020-02-03
printf() can not handle a argument of type va_list.

BlueAndi:
This bugfix is for AsyncWebSocket::printf() and AsyncWebSocket::printf_P().


sfranzyshen:
Thank you @BlueAndi !!

PR #737, opened 2020-04-05
Fix template processor corrupts (Fixes #644)

### LINKED ISSUE #644: Using template processor corrupts response HTML if it includes %
I got this library successfully serving index.htm over SPIFFS, using the example provided in the readme: `request->send(SPIFFS, "/www/index.htm", "text/html");`

Then I tried to use the template processor to alter some variables inside the HTML: `request->send(SPIFFS, "/www/index.htm", String(), false, processor);`

With the processor from the readme: 
````
String processor(const String& var)
{
  if(var == "HELLO_FROM_TEMPLATE")
    return F("Hello world!");
  return String();
}
````

This correctly replaces the `%HELLO_FROM_TEMPLATE%` with `Hello world!`, but it appears to corrupt the output HTML as well.  For example, the submit button is now missing, as well as some CSS.

I suspect this is related to some of the inline styles I wrote, which use %, such as: `style='width: 100%;height:80%'`

However, I wouldn't expect this behavior, since the template processor should just pass through any string that doesn't match, right?

... As I write that, it becomes obvious to me that the example provided in the readme returns an empty String if `var` doesn't match the template string.  So anything "surrounded" with % signs gets gobbled up.

Changing this to `return var;` recovers some of the expected HTML, but not everything. Since the %'s on both sides get stripped, the `var` variable only contains what was between them.  Trying `return "%" + var + "%";` actually seems to reset the ESP32 (triggering a RTOS watchdog).

Is there an easier way for me to ensure unexpected parts of my HTML are not altered? Maybe we can clarify that in the README?

[Example code](https://github.com/rorosaurus/captive-wifi-remote/tree/738f0564a536397dd61fe2ed70fd2d4891de897d/captive-wifi-remote)

MrKalach:


masman69:
I am curious if this is still being merged.  I am interested as well.  I would further that a change for any whitespace character \t \r \n or actual space should indicate that a percent isn't part of a template parameter.

PR #747, opened 2020-04-13
Fix LinkedList namespace to avoid redefinition with other libraries.

anmaped:
LinkedList is a custom implementation. It'll affect other libraries. I just added it to a new namespace to avoid further issues.


paclema:
Can this be merged?

PR #783, opened 2020-06-05
Added Webdav HTTP verbs

mabau:
support implementation of a WebDav server by adding additional WebDav methods


rostwolke:
This is a duplicate of PR 676

PR #821, opened 2020-08-06
Add client queue length method

cgmckeever:
A lot of the diff is trailing whitespace removal. 

Only change is 

```
size_t queueLength() { return _messageQueue.length(); }
```

Which can be used to determine current Queue

```
    AsyncWebSocket::AsyncWebSocketClientLinkedList clients = webSocket.getClients();
    size_t queue = 0;
    for(const auto& c: clients){
        queue += c->queueLength();
    }
```


rprediger:
It will be great when this method is implemented. I already use it in my projects. Here is my support for this to happen as soon as possible.

PR #851, opened 2020-09-23
fix curl command upload error

moononournation:


moononournation:
The original ESP32 FSBrowser can use curl upload SPIFFS files:
for file in `ls -A1`; do curl -F "file=@$PWD/$file" fsbrowserPlus.local/edit; done

This pull request help to prepend "/" to the filename for fulfilling the SPIFFS filename format and return HTTP 200 after finished. (avoiding curl received HTTP 500 at last)

Issue #884, opened 2020-11-16
Concurrent access of AsyncEventSourceClient::_messageQueue causes memory access violation


ul-gh:
Hi,

in contrast to any normal AsyncWebHandler derivative which runs in the async_tcp task only:

When using an AsyncEventSource instance on ESP32 platform, the AsyncEventSourceClient::_messageQueue is supposed to be accessed by both the AsyncEventSource callbacks running in async_tcp task and any application task calling AsyncEventsource::send().

This causes dereferencing of invalid pointers in AsyncEventSourceClient::_runQueue() when the linked list is iterated over while being changed from another task. Although this might happen infrequently, this is a crash/DOS/safety issue.

This was already adressed by Arjan Filius by adding mutex locking of access to the AsyncEventSource _messageQueue (see iafilius/ESPAsyncWebServer/commit/d924de1 etc.)

With this report I am suggesting a pull request introducing the mutex lock wherever the linked list is iterated over, with slightly changed position as in ArFi commits.

Plus, the same locking mechanism is introduced where the AsyncEventSource::_clients list is iterated over as the public API is also supposed to be called from different tasks.

ul-gh:
Will there be any official fix for this?

brunojoyal:
> Will there be any official fix for this?

I would also like to know this... Thank you

PR #886, opened 2020-11-16
Fix Issue #884, add mutex lock to linked list access in AsyncEventSource.cpp

ul-gh:
Now with added version for ESP8266


ul-gh:
Fix #884

ul-gh:
Cancelling this as PR 887 is the same but rebased into one single commit

PR #887, opened 2020-11-16
Fix #884, protect list concurrent access with mutex

### LINKED ISSUE #884: Concurrent access of AsyncEventSourceClient::_messageQueue causes memory access violation

ul-gh:
Same as PR 886, fix #884 but remastered into single commit.

### DIFF
--- a/src/AsyncEventSource.cpp
+++ b/src/AsyncEventSource.cpp
@@ -173,7 +173,9 @@ AsyncEventSourceClient::AsyncEventSourceClient(AsyncWebServerRequest *request, A
 }
 
 AsyncEventSourceClient::~AsyncEventSourceClient(){
-   _messageQueue.free();
+  _lockmq.lock();
+  _messageQueue.free();
+  _lockmq.unlock();
   close();
 }
 
@@ -184,32 +186,40 @@ void AsyncEventSourceClient::_queueMessage(AsyncEventSourceMessage *dataMessage)
     delete dataMessage;
     return;
   }
+  //length() is not thread-safe, thus acquiring the lock before this call..
+  _lockmq.lock();
   if(_messageQueue.length() >= SSE_MAX_QUEUED_MESSAGES){
-      ets_printf("ERROR: Too many messages queued\n");
-      delete dataMessage;
+    ets_printf("ERROR: Too many messages queued\n");
+    delete dataMessage;
   } else {
-      _messageQueue.add(dataMessage);
+    _messageQueue.add(dataMessage);
+    // runqueue trigger when new messages added
+    if(_client->canSend()) {
+      _runQueue();
+    }
   }
-  if(_client->canSend())
-    _runQueue();
+  _lockmq.unlock();
 }
 
 void AsyncEventSourceClient::_onAck(size_t len, uint32_t time){
+  // Same here, acquiring the lock early
+  _lockmq.lock();
   while(len && !_messageQueue.isEmpty()){
     len = _messageQueue.front()->ack(len, time);
     if(_messageQueue.front()->finished())
       _messageQueue.remove(_messageQueue.front());
   }
-
   _runQueue();
+  _lockmq.unlock();
 }
 
 void AsyncEventSourceClient::_onPoll(){
+  _lockmq.lock();
   if(!_messageQueue.isEmpty()){
     _runQueue();
   }
+  _lockmq.unlock();
 }
-
 
 void AsyncEventSourceClient::_onTimeout(uint32_t time __attribute__((unused))){
   _client->close(true);
@@ -225,7 +235,7 @@ void AsyncEventSourceClient::close(){
     _client->close();
 }
 
-void AsyncEventSourceClient::write(const char * message, size_t len){
+void AsyncEventSourceClient::_write(const char * message, size_t len){
   _queueMessage(new AsyncEventSourceMessage(message, len));
 }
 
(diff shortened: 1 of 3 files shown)


src/AsyncWebSynchronization.h
```diff
@@ -7,6 +7,36 @@
...
+// added unimplemented version for ESP8266
+class AsyncPlainLock
+{
+private:
+  SemaphoreHandle_t _lock;
+  mutable void *_lockedBy;
+
+public:
+  AsyncPlainLock() {
+    _lock = xSemaphoreCreateBinary();
```
BlueAndi, line 22: xSemaphoreCreateBinary() can fail in case of less memory.

src/AsyncWebSynchronization.h
```diff
@@ -7,6 +7,36 @@
...
+    xSemaphoreGive(_lock);
+  }
+
+  ~AsyncPlainLock() {
+    vSemaphoreDelete(_lock);
+  }
+
+  bool lock() const {
+      xSemaphoreTake(_lock, portMAX_DELAY);
```
BlueAndi, line 31: xSemaphoreTake() can fail

src/AsyncWebSynchronization.h
```diff
@@ -7,6 +7,36 @@
...
+      return true;
+  }
+
+  void unlock() const {
+    xSemaphoreGive(_lock);
+  }
+};
+
 // This is the ESP32 version of the Sync Lock, using the FreeRTOS Semaphore
```
BlueAndi, line 40: Its the ESP8266 version, not ESP32 ;-)

src/AsyncWebSynchronization.h
ul-gh:
```suggestion
    _lock = xSemaphoreCreateBinary();
    // If this fails, the system is likely so much out of memory that we should abort
    assert(_lock);
```

src/AsyncWebSynchronization.h
ul-gh: No, actually in this case, xSemaphoreTake, which is an alias to xQueueGenericReceive, would only ever have a nonzero return value if an effective timeout was supplied as an argument. But the portMAX_DELAY in this case means the timeout is disabled. xSemaphoreTake does either lock indefinitely or return zero.

src/AsyncWebSynchronization.h
ul-gh: No, the define is OK and for ESP8266, but still I noticed a bug there, as I accidentally left the "mutable void *_lockedBy;" member in for the supposedly "plain" lock class.

src/AsyncWebSynchronization.h
```diff
@@ -7,6 +7,36 @@
...
 
+// This is the ESP32 version of the Sync Lock, using the FreeRTOS Semaphore
+// Modified 'AsyncWebLock' to just only use mutex since pxCurrentTCB is not
+// always available. According to example by Arjan Filius, changed name,
+// added unimplemented version for ESP8266
+class AsyncPlainLock
+{
+private:
+  SemaphoreHandle_t _lock;
+  mutable void *_lockedBy;
```
ul-gh, line 18: This was supposed to be left out.. 
```suggestion
```

src/AsyncWebSynchronization.h
ul-gh: The original implementation of the AsyncWebLock (below the new class) does the same without an assert etc.

What would you suggest, put an abort() or assert() there? If yes, this should go in both places.

src/AsyncWebSynchronization.h
ul-gh: xQueueGenericReceive, I meant it only returns "pdFALSE" when invoked with a non-infinite timeout.

src/AsyncWebSynchronization.h
BlueAndi: assert() is fine and give the developer enough information I think.

src/AsyncWebSynchronization.h
BlueAndi: Yes, you are right. I expected a ```(void)xSemaphoreTake(...)```, but this may depend on the coding style.

src/AsyncWebSynchronization.h
BlueAndi: Oh yes, your right here. Looks like I read the comment and scrolled to fast down.

src/AsyncEventSource.cpp
```diff
@@ -276,56 +294,70 @@ void AsyncEventSource::_addClient(AsyncEventSourceClient * client){
     client->write((const char *)temp, 2053);
     free(temp);
   }*/
-  
+  _client_queue_lock.lock();
```
avillacis, line 297: By calling the connect callback with the client queue lock held, this code introduces a self-deadlock if the connect callback, in turn, tries to broadcast a message to the event source using AsyncEventSource::send (or tries to use any other public API that tries to take that very same lock).

I had this exact scenario in my project, and I had to fix it as shown in commit https://github.com/yubox-node-org/ESPAsyncWebServer/commit/646eab7bd8761a7ff89c627e37c0f27c49cef32f of yuboxfixes branch of the ESPAsyncWebServer fork at yubox-node-org.

ul-gh: I just now realized that the extern void *pxCurrentTCB member of class AsyncWebLock does /not/ point to the TCP TCB struct (TCB as in "transfer control block") but to the FreeRTOS stack pointer, and that this is how the AsyncWebLockGuard works..
I only did minimal testing, but I assume this is safe to pull into this PR for completeness.

I hope that the original author does find some time in the near future to go through all these forks and PRs...
Thank you very much for the notice!

PR #889, opened 2020-11-18
Fix #888, AsyncEventSourceMessage::send()

### LINKED ISSUE #888: AsyncEventSourceMessage::send() fails when message size > TCP_SND_BUF
Hi,

In file AsyncEventSource.cpp: When the SSE message size is greater than the size reported by AsyncClient::space(), which is the LWIP TCP send buffer space (CONFIG_LWIP_TCP_SND_BUF_DEFAULT, set to 5744 bytes), the event message is not sent correctly.

In function AsyncEventSourceMessage::send(), the call to "client->add()" is made using the "_data" message buffer.
While "client->add()" has a maximum message size limit of TCP_SND_BUF size, this is only added to the "_sent" bytes data member, but that property is never evaluated. There is no offset added to the _data buffer index, nor is the _data buffer ever consumed. (Until deletion, that is).

From the code:
```size_t AsyncEventSourceMessage::send(AsyncClient *client) {
  const size_t len = _len - _sent;
  if(client->space() < len){
    return 0;
  }
  size_t sent = client->add((const char *)_data, len);  <====== No offset, always starts from beginning of buffer
  if(client->canSend())
    client->send();
  _sent += sent;
  return sent; 
}
```
This simply does not work.

The following code does, this was just now tested with SSE messages of 16 kB size:

```
// This could also return void as the return value is not used.
// Leaving as-is for compatibility...
size_t AsyncEventSourceMessage::send(AsyncClient *client) {
    if (_sent >= _len) {
      return 0;
    }
    const size_t len_to_send = _len - _sent;
    auto position = reinterpret_cast<const char*>(_data + _sent * sizeof(*_data));
    const size_t sent_now = client->write(position, len_to_send);
    _sent += sent_now;
    return sent_now;
}
```
PR follows; please let me know if there are any issues.

ul-gh:
Tested with 16 kB message size: Respect number of already sent bytes
in successive calls to AsyncEventSourceMessage::send()

### DIFF
--- a/src/AsyncEventSource.cpp
+++ b/src/AsyncEventSource.cpp
@@ -137,16 +137,17 @@ size_t AsyncEventSourceMessage::ack(size_t len, uint32_t time) {
   return 0;
 }
 
+// This could also return void as the return value is not used.
+// Leaving as-is for compatibility...
 size_t AsyncEventSourceMessage::send(AsyncClient *client) {
-  const size_t len = _len - _sent;
-  if(client->space() < len){
-    return 0;
-  }
-  size_t sent = client->add((const char *)_data, len);
-  if(client->canSend())
-    client->send();
-  _sent += sent;
-  return sent; 
+    if (_sent >= _len) {
+      return 0;
+    }
+    const size_t len_to_send = _len - _sent;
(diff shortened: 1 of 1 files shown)


ul-gh:
Pushed again because this somehow did not link to the issue..

PR #922, opened 2021-01-18
Redefine DYNAMIC_JSON_DOCUMENT_SIZE if not defined

paclema:
By default the json doc size for AsyncJson is #define DYNAMIC_JSON_DOCUMENT_SIZE  1024.
When I want to handle a POST request with a bigger json, the AsyncWebServer is not able to handle it and I get an Http failure response from the web browser. But with this fix you can predefine a bigger json document size.


cpainchaud:
some please push !!!!

PR #952, opened 2021-03-07
Make LinkedList iterator safe against removal of cur item

### LINKED ISSUE #951: Use-after-free in AsyncWebServerRequest::_removeNotInterestingHeaders
Hi,
While running my code with Valgrind in an effort to troubleshoot memory-corruption issues I noticed there's a use-after-free error in this library's code.

Here the function iterates over a linked list of headers, and removes not-interesting ones from the list.
https://github.com/me-no-dev/ESPAsyncWebServer/blob/master/src/WebRequest.cpp#L183-L186

Here, the linked list's remove method correctly updates the next pointer and frees the removed item:

https://github.com/me-no-dev/ESPAsyncWebServer/blob/master/src/StringArray.h#L120-L132

However, if an item is deleted during iteration, the iterator still holds a reference to its memory, and it still uses it to retrieve the next pointer:

https://github.com/me-no-dev/ESPAsyncWebServer/blob/master/src/StringArray.h#L53

While this works in most cases and particularly on the ESP MCUs, on my testing setup I get a nice SIGSEGV.

A potential fix could be retrieving the `next` pointer ahead of time and storing it in the iterator. Then in `operator++` replace it with the "next-next" pointer and return the stored `next` pointer. This should make deleting the current item safe.

---
<small>
(If you're wondering how the hell I'm running it with valgrind, I stubbed all the Arduino calls and patched ESPAsyncTCP to use the POSIX socket API, and I'm running my application as a regular Linux program)
</small>

Depau:
This PR implements my suggested solution for #951.

Closes #951.

### DIFF
--- a/src/StringArray.h
+++ b/src/StringArray.h
@@ -47,10 +47,19 @@ class LinkedList {
 
     class Iterator {
       ItemType* _node;
+      ItemType* _nextNode = nullptr;
     public:
-      Iterator(ItemType* current = nullptr) : _node(current) {}
+      Iterator(ItemType* current = nullptr) : _node(current) {
+          if (_node != nullptr) {
+              _nextNode = current->next;
+          }
+      }
       Iterator(const Iterator& i) : _node(i._node) {}
-      Iterator& operator ++() { _node = _node->next; return *this; }
+      Iterator& operator ++() {
+          _node = _nextNode;
+          _nextNode = _node != nullptr ? _node->next : nullptr;
+          return *this;
+      }
       bool operator != (const Iterator& i) const { return _node != i._node; }
       const T& operator * () const { return _node->value(); }
       const T* operator -> () const { return &_node->value(); }


sieren:
This is great! Can we get this merged already or is this repo no longer maintained?
Seems like the issue described here happens a lot more often when compiling with current masters of ardiuno-esp32 and ESP-IDF 4.4 (Crashes on access, no matter what - verified fixed with this PR).

Depau:
FWIW I'm keeping a branch with all the bugfixes I need for my project here: https://github.com/Depau/ESPAsyncWebServer/tree/wi-se-patches

Feel free to use it until the fixes are merged.

fcapano:
This issue has had me debugging memory corruption on an ESP32 until I came across this bug report.
The fix is working nicely for me. Too bad after more than 4 months it still hasn't been merged

Depau:
> is this fix related to this issue? #1201

Try the patches and see if it fixes your issues, I can't telepathically connect to your ESP32 to see it fixes the problem. https://github.com/Depau/ESPAsyncWebServer/tree/wi-se-patches

Depau:
@teknoel since you mention websockets it might also be this other bug #953

PR #1050, opened 2021-08-29
filesystem based webserver, easy use: all on one tab

holgerlembke:


examples/simplefilewebserver/webserver.ino
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  if (LittleFS.exists(uri)) {
+    request->send(LittleFS, uri, String(), false);
+  } else {
+    notFound404(request);
+  }
+
+  delay(1);
```
Pablo2048, line 77: Very bad! - see the readme description (avoid delay() in callbacks...

examples/simplefilewebserver/webserver.ino
```diff
@@ -0,0 +1,78 @@
...
+  // default name
+  if (uri == "/") {
+    uri = webserverDefaultname;
+  }
+
+  // Are we allowd to send compressed data?
+  if (request->hasHeader("ACCEPT-ENCODING")) {
+    if (request->header("ACCEPT-ENCODING").indexOf("gzip") != -1) {
+      // gzip version exists?
+      if (LittleFS.exists(uri + ".gz"))  {
```
Pablo2048, line 65: No need to doing like this - the server can server static files internally (even with .gz content) see https://github.com/me-no-dev/ESPAsyncWebServer#serving-static-files

examples/simplefilewebserver/webserver.ino
```diff
@@ -0,0 +1,78 @@
...
+  request->send(404, "text/plain", message);
+}
+
+//*****************************************************************************************************
+void notFound(AsyncWebServerRequest *request) {
+  String uri = request->url();
+
+  // default name
+  if (uri == "/") {
+    uri = webserverDefaultname;
```
Pablo2048, line 58: No need to do it like this - use the setDefaultFile() method instead...

Pablo2048:
Please see my comments...

PR #1081, opened 2021-10-30
Publish to LibraryManager (fix library.properties)

ricardojlrufino:


library.properties
```diff
@@ -1,9 +1,9 @@
-name=ESP Async WebServer
+name=ESPAsyncWebServer
```
Rotzbua, line 1: Renames the library and breaks backwards compatibility

PR #1084, opened 2021-11-04
Content type startswith

devel0:
allow canHandle not to discard request such as

```
Content-type: application/json; charset=utf-8
```

without this patch only valid request for AsyncCallbackJsonWebHandler are

```
Content-type: application/json
```

### DIFF
--- a/src/AsyncJson.h
+++ b/src/AsyncJson.h
@@ -193,6 +193,19 @@ public:
   void setMaxContentLength(int maxContentLength){ _maxContentLength = maxContentLength; }
   void onRequest(ArJsonRequestHandlerFunction fn){ _onRequest = fn; }
 
+  bool startsWithIgnoreCase(const char *a, const char *b)
+  {
+    if (a == NULL || b == NULL) return false;
+    int i = 0;
+    while (b[i])
+    {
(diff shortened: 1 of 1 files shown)


Domochip:
Hi, I've just encountered this issue.
My solution is a bit different.
I replaced : 
`if ( !request->contentType().equalsIgnoreCase(JSON_MIMETYPE) )`
by : 
`if ( !request->contentType().substring(0, strlen(JSON_MIMETYPE)).equalsIgnoreCase(JSON_MIMETYPE) )`

Simpler to read but create a little heap consumption.
What's your mind about it?

devel0:
I agree, your solution is simpler to read, while mine was designed to work with existing string.

Actually I see this PR isn't yet merged so you could make another PR mentioning this one in your message so that when one of the two will accepted the other could discarded.

Thanks for your suggestion.

RayPS:
Just encountered the same issue, given `500` error with no any clue or error message, spent lot of hours debugging this library until I found this line:
https://github.com/me-no-dev/ESPAsyncWebServer/blob/2f784268f0a358741ee6384480d48656e159d726/src/AsyncJson.h#L208

PR #1105, opened 2021-12-30
AsyncJson : Fix handling of Content-Type that end with charset

Domochip:
Like in #1084
AsyncCallbackJsonWebHandler::canHandle() doesn't work if Content-Type contains 
`application/json; charset=utf-8`
which is an eligible one for AsyncJson.

This solution is simpler to read but create a little heap consumption.
@me-no-dev : Could you choose between both solution.
Thanks you


Domochip:
up

Domochip:
up

PR #1125, opened 2022-03-03
Fix macro invocation with post-increment index as argument

NDMDevel:
In WebRequest.cpp at line 150:
`while (i<len && __is_param_char(((char*)buf)[i++]));`
the macro `__is_param_char` will expand producing multiple `i++` invocations making the algorithm to (sometimes) fail.

### DIFF
--- a/src/WebRequest.cpp
+++ b/src/WebRequest.cpp
@@ -28,7 +28,14 @@
 
 static const String SharedEmptyString = String();
 
-#define __is_param_char(c) ((c) && ((c) != '{') && ((c) != '[') && ((c) != '&') && ((c) != '='))
+bool is_param_char(char c) {
+  return 
+	(c != 0) && 
+	(c != '{') && 
+	(c != '[') && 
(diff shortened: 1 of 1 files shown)


mrengineer7777 (commented):
I agree: this looks like a bug.  "__is_param_char" macro expansion will cause multiple increments of i++.  That is a dangerous macro.  Rather than a macro, perhaps we should be using a function here?
```
bool is_param_char(char c) {
  return 
	(c) && 
	((c) != '{') && 
	((c) != '[') && 
	((c) != '&') && 
	((c) != '=');
}
```
From reading your code, it should also work around the bug, but I can't test it.

NDMDevel:
Yes, I also think is best to replace the macro with a function.
I'll test your function in my code, although is certain that will produce the same result as the macro.
Thanks

mrengineer7777 (approved):
Looks good to me

mrengineer7777:
@me-no-dev Please review.

PR #1135, opened 2022-04-04
Update README.md for ArduinoJSON 6

mrengineer7777:
ESPAsyncWebServer library supports AJ6, but the examples in the readme have not been updated.

See https://arduinojson.org/v6/doc/upgrade/.  There may be other changes necessary for AJ6.


Rotzbua (approved):

chess-levin:
I'm trying to use your simple JSON response example with ArduinoJson 6.1 and can't compile. Is there a work around for this issue?

image

mrengineer7777:
@chess-levin Correct.  You want to switch from JsonBuffer to JsonDocument.  The dynamic part has been deprecated.  DynamicJsonDocument is actually a fixed size.  The only difference between Static and Dynamic is where the buffer is allocated.

See https://arduinojson.org/ "Fixed Memory Allocation".

PR #970, opened 2021-04-19
Fix compiler error for ESP32-C3

khoih-prog:
Compile error on ESP32-C3 for even simplest example [simple_server](https://github.com/me-no-dev/ESPAsyncWebServer/tree/master/examples/simple_server)

```
/home/kh/Arduino/libraries/ESPAsyncWebServer-master/src/AsyncWebSocket.cpp: In member function 'IPAddress AsyncWebSocketClient::remoteIP()':
/home/kh/Arduino/libraries/ESPAsyncWebServer-master/src/AsyncWebSocket.cpp:843:28: error: call of overloaded 'IPAddress(unsigned int)' is ambiguous
         return IPAddress(0U);
                            ^
```

PR #1142, opened 2022-04-26
Update AsyncWebSocket.cpp

har-in-air:
Fix compile error for ESP32-C3 RISC-V toolchain ( ambiguous type)

Original compile error : 

xxx/ESPAsyncWebServer/src/AsyncWebSocket.cpp:832:28: error: call of overloaded 'IPAddress(unsigned int)' is ambiguous
return IPAddress(0U);
^

xxx/.arduino15/packages/esp32/hardware/esp32/2.0.1/cores/esp32/IPAddress.h:51:5: note: candidate: 'IPAddress::IPAddress(const uint8_t*)'
IPAddress(const uint8_t *address);
^~~~~~~~~
xxx/.arduino15/packages/esp32/hardware/esp32/2.0.1/cores/esp32/IPAddress.h:50:5: note: candidate: 'IPAddress::IPAddress(uint32_t)'
IPAddress(uint32_t address);

### DIFF
--- a/src/AsyncWebSocket.cpp
+++ b/src/AsyncWebSocket.cpp
@@ -829,7 +829,7 @@ void AsyncWebSocketClient::binary(AsyncWebSocketMessageBuffer * buffer)
 
 IPAddress AsyncWebSocketClient::remoteIP() {
     if(!_client) {
-        return IPAddress(0U);
+        return IPAddress((uint32_t)0);
     }
     return _client->remoteIP();
 }


aaroneiche:
I wanted to push this as well. I was unable to get the most recent release to work with my ESP32-C3 project, and this branch compiled without issue.

quiret:
Thanks.

FedericoBusero:
Please rename the subject of this issue to "Fix compilation error on RISC-V processor ESP32-C3"

bluet:
related https://github.com/me-no-dev/ESPAsyncWebServer/pull/970

src/AsyncWebSocket.cpp
```diff
@@ -829,7 +829,7 @@ void AsyncWebSocketClient::binary(AsyncWebSocketMessageBuffer * buffer)
 
 IPAddress AsyncWebSocketClient::remoteIP() {
     if(!_client) {
-        return IPAddress(0U);
+        return IPAddress((uint32_t)0);
```
willsmart, line 832: It'd be great to get this merged some day. There seems to be a whole history of people forking the lib to fix this one line!
I'm using current libs and modern platformio with a ESP32 chip and this error is a hard blocker for using this library at all.

The overload with no arguments defaults to a zero address, so I'd suggest leaving the address out altogether...
```suggestion
        return IPAddress();
```
Both work though, so 🤷

PR #1191, opened 2022-08-04
Add example for a simple WiFi Manager

linusrj:
This example is currently only tested for ESP32, however it can not be too hard to fix compatibility with ESP8266.


sebosfato:
i tried but there is no preference file

Issue #1210, opened 2022-09-17
SPIFFSEditor not displaying file contents on ESP32 with LittleFS


steff393:
Hi, 
thanks for this great library. I used it a lot on ESP8266. But now with ESP32 and LittleFS I found that the SPIFFSEditor doesn't display the file contents (only the file names on the left frame).
I got the error message:
[E][vfs_api.cpp:29] open(): myFile.txt does not start with /

The following change (i.e. adding the "/" in front of the filename) fixes this behavior:
https://github.com/me-no-dev/ESPAsyncWebServer/blob/f71e3d427b5be9791a8a2c93cf8079792c3a9a26/src/SPIFFSEditor.cpp#L397
`request->_tempFile = _fs.open(String("/") + request->arg("edit"), "r");`

But I'm not sure if this is really the solution or only a quick workaround. What is your opinion?

blazoncek:
I have noticed the same behaviour, though the fork we use (from @Aircoookie) works correctly on ESP32 but fails on ESP32-S2 and ESP32-S3.

slimline33:
The same problem here on my esp32 wroom.

![image](https://user-images.githubusercontent.com/3323812/192561509-343ff813-c560-4180-89fe-fa0d23f10048.png)

the workaround does not help :(

steff393:
After I had to introduce that "/" into multiple different places, I searched for a more generic approach: 
It now works also, when I introduce a
```
#ifdef ESP32
        output += "/";
#endif
```

here in between line 465 and 466:
https://github.com/me-no-dev/ESPAsyncWebServer/blob/f71e3d427b5be9791a8a2c93cf8079792c3a9a26/src/SPIFFSEditor.cpp#L464-L468

This makes the filenames in the left frame of the SPIFFSEditor being displayed as "/index.html" instead of "index.html" and also the subsequent actions like Save, Delete, etc. use then the correct filename including "/".

So from my point of view this could be a real solution.

However I don't yet understand, why it's working in Aircookies fork. So there might be also another solution.

blazoncek:
@steff393 You may want to use '/' instead of "/".

Aircoookie:
Thank you for finding what seems like a good solution!
Note that at least for my fork/environment, filenames correctly include `/` as is, so your proposed change makes it `//`, which appears to work fine, but feels hacky.

[...]

PR #1220, opened 2022-10-10
ESP32: Use path() instead of name() to keep '/' before filename (#1210)

steff393:
The leading slashes were missing in SPIFFSEditor file overview. Therefore any requests to open, save, delete, etc. didn't work (seems to be introduced with new LittleFS implementation on current Arduino core).
Using `path()` instead of `name()` keeps the slash in front of the filename and fixes the errors.

See additional details in #1210.


danielkucera:
@me-no-dev I have exactly the same problem, this patch fixes it.
Can you please merge?

Issue #1249, opened 2022-12-04
Poor selection of % as placeholder delimiter because is commonly used in CSS and JavaScript code.


aiotech-pub:
% is used commonly in CSS to define style and aspect ratio of elements related to the parent. It is also the modulus operator in JavaScript. This means that any couple of % inside a String or HTML file where is present at least one place holder can not have any other % inside because they will be wrongly interpreted as the delimiters of a placeholder. In my opinion a single character can not be used as delimiter so I strongly suggest a couple of characters as may be %%.

aiotech-pub:
At the moment to quickly solve my problem I replaced % with _ and everything works now.

Sveninndh:
I ran into the same problem, it took me a few hours. When studying the source code, I noticed two things that are unfortunately not documented.
1.) The maximum length of a placeholder is limited to 32 bytes.
     Don't use more as it could cause memory issues.
2.) If you use % within a string or HTML file,
     you must escape all occurrences of % - which are not enclose a placeholder  - to %%.
     Example: 'Width: 50%;' needs to be changed to 'width: 50%%;'

hmueller01:
@Sveninndh Thanks for your posting. This helped me a lot! I would suggest opening a PR to add this to the README.md but I doubt it will be merged as @me-no-dev didn't do anything since a year ...

DeeEmm:
This has been mentioned several times before. I also reported this issue just a short time ago...

https://github.com/me-no-dev/ESPAsyncWebServer/issues/1234

Escaping the % characters is little more than a hack as it generates non-compliant css... any solution that creates another problem is not a solution at all ;)

Sveninndh:
@DeeEmm Why do you think it would generates non-compliant css?

After processing the callback template processor the result should be compliant css.
Or do I miss something?

DeeEmm:
Incorrect source syntax. Try it in a validator. 

% is not a valid escape character for CSS

W3C recommends that it is "always preferable to display characters in their original form" because "using escapes can make it difficult to read and maintain **source** code, and can also significantly increase file size."

Compliant code is more than what you send to the browser. I mean, lets just obscure it by using CSS escapes and Gzipping it, after all if it is non-human readable no one will know, right?

Semantics are very important. Remember IE6??

Coding standards exist for a reason.

/DM

Sveninndh:
Some valid points.
try my fix in the attachment and tell me if it works.
[...]

Issue #1234, opened 2022-11-16
Template processor corrupts output


DeeEmm:
##Outline of problem

I am using ESPAsyncWebserver to serve webpages that contain sensor data.

I have a single monolithic minified file that contains the HTML  + CSS + Javascript components of the web page. This single file structure is a necessity determined by space constraints.

My setup is working but I have decided to move from websockets to templating to address some unrelated interrupt issues.

Enabling the template function results in the % characters being stripped from the output. This has already been reported in bug #644 and a fix proposed with PR #737. However this issue is still present and remains unaddressed. The PR as never merged.

Several solutions were proposed within #644 , but none of them are working in my case...

1- Delimiting the % character with %%. Whilst I disagree the delimiting the % within the page by using %% is a valid solution, I did try it and found that it does not actually work as it simply creates invalid CSS.

2- Changing the delimiter. I tried changing the delimiter by defining an alternate TEMPLATE_PLACEHOLDER character that is unused elsewhere in my code '$', but this still strips out the % characters in addition to the alternate character.

3- PR #737 - I manually added the PR code to test this out, however it also fails to address the issue and the output is still corrupted. It also does not work even in conjunction with solutions 1 and 2. Even if this did work it's not really a manageable solution as my project code is in the public domain so I do not want to have to support customised libraries on top of supporting the project itself. My preference is to use vanilla libraries to avoid complexity

So I am kind of stuck needing to get this to work. I cannot use websockets and I don't really want to have to resort to writing a templating engine as this is basically doubling up on what should already work.

What's the status of this project? Is a fix for this issue even likely given that the previous issue's fix was not integrated into the main code?

Just in case it's a dumb issue of my own creation, here's a sample of my code...

index.html
```
    <!-- Status Modal -->
    <div id="statusModal" class="modal">
      <div class="modal-content">
        <span class="closeStatusModal">&times;</span>
          <h4>Storage</h4>
          <hr>
          <div class="config-row"><label class="config-label">Memory Size:  </label><span class="config-value" id="xSPIFFS_MEM_SIZE">%SPIFFS_MEM_SIZE%</span> (bytes)</div>
          <div class="config-row"><label class="config-label">Memory Used:  </label><span class="config-value" id="xSPIFFS_MEM_USED">%SPIFFS_MEM_USED%</span> (bytes)</div>
      </div>        
    </div>
```

webserver.cpp
```
...
      if (SPIFFS.exists("/index.html")) {
[...]

PR #1260, opened 2023-01-03
Improvement of template processing (see issues #1234 and #1249)

Sveninndh:
AsyncAbstractResponse::_fillBufferAndProcessTemplates()

1.) A parameter is enclosed by the % delimiter.
2.) Parameter names can have a maximum of 32 characters.
3.) Parameter names cannot contain illegal characters. (\ .:<>{}',;"/)

A parameter is only recognized and replaced if all of these conditions are met.
Otherwise the text source code is not changed.
Using double % is no longer necessary.
Nevertheless, this mechanism will continue to be supported for reasons of compatibility.

### DIFF
--- a/src/WebResponses.cpp
+++ b/src/WebResponses.cpp
@@ -383,6 +383,7 @@ size_t AsyncAbstractResponse::_fillBufferAndProcessTemplates(uint8_t* data, size
   if(!_callback)
     return _fillBuffer(data, len);
 
+  const char* illegalChars = " .:<>{}/',;\"\\";
   const size_t originalLen = len;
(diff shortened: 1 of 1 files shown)


Jakeduncan00:
1.) A parameter is enclosed by the % delimiter.
2.) Parameter names can have a maximum of 32 characters.
3.) Parameter names cannot contain illegal characters. (\ .:<>{}',;"/)

A parameter is only recognized and replaced if all of these conditions are met.
Otherwise the text source code is not changed.
Using double % is no longer necessary.
Nevertheless, this mechanism will continue to be supported for reasons of compatibility.

alessioburgassi (approved):

PR #1274, opened 2023-01-31
add support for brotli compressed files

vortigont:
static files could be compressed with brotli and must have '.br' extension.
It works same way as with .gz files, brotli has preference if both file types are present

### DIFF
--- a/src/ESPAsyncWebServer.h
+++ b/src/ESPAsyncWebServer.h
@@ -46,6 +46,14 @@
 
 #define DEBUGF(...) //Serial.printf(__VA_ARGS__)
 
+// const strings
+static const char c_gzip[] = "gzip";
+static const char c_brotli[] = "br";
+static const char c_gz[] = ".gz";
+static const char c_br[] = ".br";
+static const char c_cencoding[] = "Content-Encoding";
+static const char c_cdisposition[] = "Content-Disposition";
+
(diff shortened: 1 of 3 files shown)


mathieucarbou:
Just bumped into that. I had the issue recently on a project.

Brotli support sadly is not implemented the same way on all browsers and most of them (all chromium based browser and firefox) only allow it over https connection... 

https://firt.dev/brotli-https/

vortigont:
true. Chrome violates standards here, it intentionally ignores `content-encoding` header with Brotli over plain http just for the sake of some possible issues with ancient proxies or so. IMO it's wrong by design, but google does not care really, they just dominate the Web.

mathieucarbou:
> true. Chrome violates standards here, it intentionally ignores `content-encoding` header with Brotli over plain http just for the sake of some possible issues with ancient proxies or so. IMO it's wrong by design, but google does not care really, they just dominate the Web.

Well, its Chromium, really, so consequently, also Microsoft browser is impacted.
And oddly, Mozilla has decided to do the same also.
Pretty sure there are some political reasons more than technical ones.

vortigont:
For the time when I created this PR Firefox was working fine with .br via plain HTTP, so that I tested it. But when I found out the all Chromium-based family browsers can't work with it, I've cancelled the PR. So sad :(

PR #1276, opened 2023-02-02
set real "Last-Modified" header based on file's LastWrite time

vortigont:
Get file's LastWrite timestamp for file handlers (if supported by FS driver) and construct proper `Last-Modified` header. Works fine for LittleFS. If not supported by FS than fallback to previous implementation with manual setup for `Last-Modified` value.
This way caching in browsers really works as it should be for any file updates on fs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -823,8 +823,11 @@ handler->setCacheControl("max-age=30");
 ```
 
 ### Specifying Date-Modified header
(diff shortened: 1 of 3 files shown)


vortigont:
merged in https://github.com/mathieucarbou/ESPAsyncWebServer

PR #1301, opened 2023-04-12
Fix RFC 7230 violation

zalupoi:
Spaces can not be added to chunk sizes. Add leading zeros instead.
See https://github.com/nodejs/node/issues/47528 for more details.


Jaybee2020:
Thanks @zalupoi , thanks nthd!

PR #1351, opened 2023-10-16
Fix build on Arduino ESP32 v3.0.0 and IDF 5.1.1

blackhack:
This fixes compilation on the latest Git version of the Arduino ESP32 core due to API changes in IDF 5.1.

Of course, these changes make ESPAsyncWebServer incompatible with prior Arduino core versions.


ayushsharma82:
@blackhack Any way we can use a compiler flag to detect old ESP32 arduino core? I think we should look into making it backwards-compatible.

PR #1362, opened 2023-11-16
data type typecasting error

VinkIoT:
at line no. 832 IPAddress function argument data type typecasting to uint32_t


haennes:
Same problem here. Idependently discovered this fix and it "just works".

PR #1436, opened 2024-10-02
Chunked Response broken on Safar/iOS

dl9rdz:
Chunked responses cause Safari on iOS 18 to display "Safari can't open the page. The error was "cannot parse response"".

The reason is that the handling of chunked response is not standard conforming in the current implementation. In contrast to a comment in the source code, the standard (RFC2616) does not allow spaces between the chunk size and the crlf. Most other browsers are more tolerant in this regard, but not Safari on iOS.

There is an easy fix: As also stated in the existing comment, leading zeros are standard compliant. So it is possible to generate an equivalent, standard compliant chunked encoding that is also correctly handled by iOS by extending the size prefix not with spaces at the end, but with leading zeros.


Sojourneer:
I'm getting the same problem with node.js http-proxy: Illegal character in chunk size.

PR #1452, opened 2024-12-12
Fix: Pointer was used after removed

### LINKED ISSUE #951: Use-after-free in AsyncWebServerRequest::_removeNotInterestingHeaders

controlol:
Implements the solution by @A40in in issue #951 

Closes #951

### DIFF
--- a/src/WebRequest.cpp
+++ b/src/WebRequest.cpp
@@ -180,11 +180,11 @@ void AsyncWebServerRequest::_onData(void *buf, size_t len){
 
 void AsyncWebServerRequest::_removeNotInterestingHeaders(){
   if (_interestingHeaders.containsIgnoreCase("ANY")) return; // nothing to do
-  for(const auto& header: _headers){
-      if(!_interestingHeaders.containsIgnoreCase(header->name().c_str())){
-        _headers.remove(header);
-      }
-  }
+  StringArray *a = &_interestingHeaders;
+  auto cmpr = LinkedList<AsyncWebHeader*>::Predicate{[&a](AsyncWebHeader* h) {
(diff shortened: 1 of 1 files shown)


PhLuReh:
Applying this PR as [patch](https://[patch-diff.githubusercontent.com/raw/me-no-dev/ESPAsyncWebServer/pull/1452.patch](https://patch-diff.githubusercontent.com/raw/me-no-dev/ESPAsyncWebServer/pull/1452.patch)) fully solved the issue.

